PR #153, opened 2023-03-01
Change 'Mod' -> 'Mode' in english launch options

legoandmars:
On the "Launch" page, there are currently three launch options:
- `Oculus Mod`
- `Desktop Mod`
- `Debug Mod`

In my opinion, these should be translated to "Mode" instead, at least in English. "Mod" implies that it's a separate third party modification, such as installing another DLL

All of these options come natively with the software (`-vrmode oculus` and `fpfc` come with the game, and `--verbose` comes with BSIPA)

### DIFF
--- a/assets/jsons/translations/en.json
+++ b/assets/jsons/translations/en.json
@@ -23,11 +23,11 @@
         },
         "version-viewer": {
             "launch-mods": {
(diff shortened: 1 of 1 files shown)


GaetanGrd (approved):
Thank you very much for your contribution!

PR #214, opened 2023-04-12
Changed from "Sie" to "Du"

XCraftTM:
The Whole Translation was written in "Du" so the new Changes need to be also written in "Du" and not in "Sie".


Zagrios (approved):

Zagrios:
Thanks!

PR #249, opened 2023-06-18
Change wording for a tad more accuracy / Fix typo.

AltyFox:
## Scope
"performances" should have been singular, but reworded to remove implication that it *will* increase performance when that is *not* the case for everyone.


Zagrios (approved):

Zagrios:
Nice, thanks for that 👌

PR #256, opened 2023-06-27
Move @types/color to devDependencies

Insprill:
## Scope

When running `npm install`, it was failing to find the `@types/color` dependency.
Moving this dependency to `devDependencies` with the others fixed the issue.

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |


Zagrios (approved):
thanks 😉

PR #257, opened 2023-06-28
Get running processes independent of platform

Insprill:
## Scope

Currently, running processes are determined by Windows registry keys, and by executing the `tasklist` command. This means that none of those functions work under Linux. This PR aims to fix that issue.

## Implementation

I've brought in the [ps-list](https://www.npmjs.com/package/ps-list) package, which is a platform-independent way of getting all running processes on the system. 

`SteamService#getActiveUser` has been removed, and `SteamService#steamRunning` now uses `ps-list` to check for any processes with `steam` in the name.

`UtilsService#taskRunning` has also been refactored to work similarly. With this refactor, it also had to be made async, bringing a few other methods along for the ride.

## How to Test

I'm not able to test this change on Windows, only Linux. Things appear to be functioning in my testing, but further testing should be done.

The easiest way to test these changes is to perform the following actions:

- Launch an instance with Steam running.
- Launch an instance with Steam closed.
- Launch an instance with another instance running.

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |


src/main/services/steam.service.ts
```diff
@@ -25,18 +26,9 @@ export class SteamService{
...
-        const keys = res?.["HKCU\\Software\\Valve\\Steam\\ActiveProcess"];
-
-        if(!keys?.exists){ throw "Key \"HKCU\\Software\\Valve\\Steam\\ActiveProcess\" not exist"; }
-
-        return (keys.values?.ActiveUser.value || undefined) as number;
-    }
-
-    public steamRunning(): Promise<boolean>{
-        return this.getActiveUser()
-            .then(userId => !!userId)
```
Zagrios, line 39: I think it should be preferable to stick with getActiveUser on windows cause some users don't have steam in there tasklist for some reason even if steam is running. And at least getActiveUser handle if the user is not logged

Zagrios (approved):
nice 🙌

PR #258, opened 2023-06-28
Properly kill the download process on non-Windows systems

Insprill:
## Scope

Currently, the `ctrlc-windows` package kills the download process when canceled.
This doesn't work on non-Windows systems, which this PR aims to fix.

## Implementation

This PR fixes the issue by only using the `ctrlc` package on Windows and using the normal `#kill` method on other operating systems, which send the appropriate `SIGTERM` signal to gracefully shutdown the process. 

## How to Test

The easiest way to test the change would be to start downloading a new version and cancel it part way through.

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |


Zagrios (approved):
thanks

PR #259, opened 2023-06-28
Use the platform native file manager when opening instance folder

Insprill:
## Scope

Currently, when attempting to open the folder an instance is installed in, it runs a Windows command to open file explorer. This PR changes the behavior to work on all platforms.

## Implementation

Instead of executing a Windows command, Electron's `shell#openPath` method is used.

## How to Test

- Open the install folder of an instance.

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |

### DIFF
--- a/src/main/ipcs/bs-version-ipcs.ts
+++ b/src/main/ipcs/bs-version-ipcs.ts
@@ -38,7 +38,6 @@ ipcMain.on('bs-version.installed-versions', async (event, req: IpcRequest<void>)
 ipcMain.on("bs-version.open-folder", async (event, req: IpcRequest<BSVersion>) => {
    const localVersionService = BSLocalVersionService.getInstance();
    const versionFolder = await localVersionService.getVersionPath(req.args);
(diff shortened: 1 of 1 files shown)


src/main/ipcs/bs-version-ipcs.ts
```diff
@@ -39,6 +39,9 @@ ipcMain.on("bs-version.open-folder", async (event, req: IpcRequest<BSVersion>) =
    const localVersionService = BSLocalVersionService.getInstance();
    const versionFolder = await localVersionService.getVersionPath(req.args);
    (await pathExist(versionFolder)) && exec(`start "" "${versionFolder}"`);
```
Zagrios, line 41: can't you delete this now ?

Zagrios, line 41: can't you delete this now ?

Zagrios (changes requested):

Zagrios (approved):
nice 🙌

PR #260, opened 2023-06-28
Use the correct steam installation path on Linux

Insprill:
## Scope

Currently, the steam installation path is retrieved from the Windows registry.
This PR allows the application to get the correct installation directory on Linux.

## Implementation

When getting the installation path, a switch statement defines the path, depending on the operating system. On Windows, the path is retrieved the same as before. On Linux, it's retrieved from `~/.steam/steam`, which, to my knowledge, Steam will always create.

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |

### DIFF
--- a/src/main/services/steam.service.ts
+++ b/src/main/services/steam.service.ts
@@ -1,11 +1,13 @@
 import { UtilsService } from "./utils.service";
 import regedit from 'regedit'
-import path from "path";
+import { path, join } from "path";
 import { parse } from "@node-steam/vdf";
 import { readFile } from "fs/promises";
 import { spawn } from "child_process";
 import { pathExist } from "../helpers/fs.helpers";
 import log from "electron-log";
+import psList from 'ps-list';
+import os from 'os';
 
 export class SteamService{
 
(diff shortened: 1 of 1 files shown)


src/main/services/steam.service.ts
```diff
@@ -42,20 +44,28 @@ export class SteamService{
...
-    let res = '';
-    if(win64.exists && win64?.values?.InstallPath?.value){ res = win64.values.InstallPath.value as string; }
-    else if(win32.exists && win32?.values?.InstallPath?.value){ res = win32.values.InstallPath.value as string; }
-    this.steamPath = res;
-    return this.steamPath;
+    if(this.steamPath){ return this.steamPath; }
+
+    switch (process.platform) {
+      case "linux":
+        this.steamPath = path.join(os.homedir(), '.steam', "steam");
```
Zagrios, line 51: you can use electron instead of "os" to get homedir 
https://stackoverflow.com/a/54387144

src/main/services/steam.service.ts
```diff
@@ -1,11 +1,13 @@
...
 import regedit from 'regedit'
-import path from "path";
+import { path, join } from "path";
 import { parse } from "@node-steam/vdf";
 import { readFile } from "fs/promises";
 import { spawn } from "child_process";
 import { pathExist } from "../helpers/fs.helpers";
 import log from "electron-log";
+import psList from 'ps-list';
+import os from 'os';
```
Zagrios, line 10: os and ps-list has not added in package.json and not needed in this branch I believe

Zagrios (changes requested):

Zagrios (approved):
nice 🙌

src/main/services/steam.service.ts
```diff
@@ -1,11 +1,12 @@
 import { UtilsService } from "./utils.service";
 import regedit from 'regedit'
-import path from "path";
+import { path, join } from "path";
```
RedlineTriad, line 3: For me importing join here caused an error when running, removing it fixed it.
Doesn't look like it's used anyways.

Insprill: Fixed in 90d4ebc7e3d5d0e3e0c33c0d651d1597dcfa3434

PR #264, opened 2023-06-28
fix(linux): search for info file case insensitively

RedlineTriad:
On linux only exactly Info.dat was matched, and many maps use different casing.

## Scope

Fixes map downloads on linux.

## Implementation

This just gets all files, and finds info.dat by comparing lowercase strings.
Slightly inefficient on windows, but I think a consistent implementation is worth it.

## How to Test

Download random maps on linux, previously that would throw errors quickly, since the casing was wrong.
Now it works without issues.

## Notes

I can remove the formatting changes, but it's a bit annoying since my IDE autoformats and I have to sleep.
I can make a PR with automatic formatting checks with prettier and eslint if that is of interest.

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |

### DIFF
--- a/src/main/helpers/fs.helpers.ts
+++ b/src/main/helpers/fs.helpers.ts
@@ -55,14 +55,11 @@ export async function getFoldersInFolder(folderPath: string, opts?: {ignoreSymli
 export async function getFilesInFolder(folderPath: string): Promise<string[]> {
     if(!(await pathExist(folderPath))){ return []; }
 
-    const files = await readdir(folderPath, {withFileTypes: true});
+    const dirEntries = await readdir(folderPath, {withFileTypes: true})
 
-    const promises = files.map(async file => {
-        if(file.isFile()){ return path.join(folderPath, file.name); }
-        return undefined;
-    });
-
-    return (await Promise.all(promises)).filter(file => file);
+    return dirEntries
+      .filter(entry => entry.isFile())
+      .map(file => path.join(folderPath, file.name));
 }
 
 export function moveFolderContent(src: string, dest: string): Observable<Progression>{
--- a/src/main/helpers/fs.helpers.ts
+++ b/src/main/helpers/fs.helpers.ts
@@ -12,7 +12,7 @@ export async function pathExist(path: string): Promise<boolean> {
     }catch(e){
         return false;
     }
-    
+
 }
 
 export async function ensureFolderExist(path: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


Zagrios (changes requested):

src/main/services/additional-content/local-maps-manager.service.ts
```diff
@@ -90,23 +90,25 @@ export class LocalMapsManagerService {
...
-        
+
         return shasum.digest("hex");
     }
 
     private async loadMapInfoFromPath(mapPath: string): Promise<BsmLocalMap>{
-        const infoFilePath = path.join(mapPath, "Info.dat");
+        const files = await getFilesInFolder(mapPath);
+        const infoFilePaths = files
+          .filter(file => (path.basename(file).toLowerCase() === "info.dat"))
```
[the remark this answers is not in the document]
RedlineTriad: Sure, not that familiar with javascript so I didn't know the best method to use.

src/main/helpers/fs.helpers.ts
```diff
@@ -52,16 +52,29 @@ export async function getFoldersInFolder(folderPath: string, opts?: {ignoreSymli
...
 
-export function moveFolderContent(src: string, dest: string): Observable<Progression>{    
-    const progress: Progression = { current: 0, total: 0 }; 
+export async function getFilesInFolder(folderPath: string): Promise<string[]> {
+    if(!(await pathExist(folderPath))){ return []; }
+
+    const files = await readdir(folderPath, {withFileTypes: true});
+
+    const promises = files.map(async file => {
+        if(file.isFile()){ return path.join(folderPath, file.name); }
```
Zagrios, line 61: The function in map don't need to be async

Zagrios: Without async, you can use the 'reduce' method to avoid two full iterations over the array by removing the filter at the end 😉

Zagrios (changes requested):

src/main/helpers/fs.helpers.ts
RedlineTriad: I copied this from FoldersInFolders, just with a flipped condition and no symlink handling.
But I guess I don't need the async function that was used there.

Would be really nice to add eslint, prettier, and maybe sonar to CI so computers can find these issues instead of you or I.
But I guess that will be it's own PR.

RedlineTriad: Fixed.

src/main/services/additional-content/local-maps-manager.service.ts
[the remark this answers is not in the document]
RedlineTriad: Fixed.

Zagrios (approved):
nice 🙌

PR #266, opened 2023-06-29
fix(linux): skip oculus library detection

RedlineTriad:
## Scope
This fixes a startup exception on linux, due to attempted windows registry read.
The combination of this and #264 makes downloading songs error free on linux.

## Implementation
Currently I just skip Oculus library detection since there are very few oculus users on linux.
However, just in case somebody wants to know why it isn't working I left a log message.

I can also add a `// TODO` or we could also open a ticket, but I am lazy.

## How to Test
Open client on linux and observe lack of exception compared to master branch.

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |

### DIFF
--- a/src/main/services/oculus.service.ts
+++ b/src/main/services/oculus.service.ts
@@ -2,6 +2,7 @@ import { UtilsService } from "./utils.service";
 import regedit from "regedit";
 import path from "path";
 import { pathExist } from "../helpers/fs.helpers";
+import log from "electron-log";
 
 export class OculusService {
     private static instance: OculusService;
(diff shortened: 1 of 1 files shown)


Zagrios (approved):

RedlineTriad:
Is this waiting on anything before it can get merged?

RedlineTriad:
Updated to fix merge conflicts after prettify.

Zagrios:
thanks 🙌

Issue #265, opened 2023-06-29
[FEAT.] : CI for automated builds, test, linting, code quality, and coverage.


RedlineTriad:
**Is your feature request related to a problem? Please describe.**
Currently it's easy for PRs to break builds, and introduce bad formatting.
This means more working during reviews, and also accidental diffs in other PRs if there is non standard formatting.

**Describe the solution you'd like**
I think it would be good to have this automated for PR's and main branch with GitHub Actions
- [x] Builds
- [x] Tests
- [x] Linting
- [x] Sonarcloud code quality
- [ ] Sonarcloud test coverage

**Describe alternatives you've considered**
Instead of Sonarcloud we could also use Codacy, or Code Climate, though I have less experience with it.

**Additional context**
Automated deployment would also be nice, but this is a relatively easy to implement start.
Also, I would be willing to implement this myself, though I might need permissions or someone else to do it for sonar.

RedlineTriad:
Mostly creating this issue to figure out if this would even be wanted or if I would be wasting my time by creating PR's.
@Zagrios So if you could tell me if you were open to some automation please give me the go-ahead :D

Zagrios:
Yeah go ahead 😉 🚀

Zagrios:
I've added SonarCloud GitHub action for code quality, but idk if it include test coverage tho 🤔

RedlineTriad:
I can try and add test coverage in a separate pr, not always easy especially with multiple platforms.

Zagrios:
As we talked here https://github.com/Zagrios/bs-manager/commit/b3251595ab167d2fc0ee07357739fff0fa938bfe#commitcomment-120406253, I think we can drop code coverage for now and keep workflows for Build, Lint, And SonarCloud auto analysis

PR #267, opened 2023-06-30
Adding CI starts work on #265

RedlineTriad:
## Scope

This starts work on #265

## Implementation

Currently just starting with builds on all 3 major OS's

## How to Test
Look at PR state :D

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |


RedlineTriad:
There seem to be a lot of typescript, eslint, and formatting issues for now, so tests and linting fails.
But I think it's a good first step to just add them for now, and see how much can be fixed.

If it turns out to be too hard to fix we can look into differential analysis so we only fail on new issues, so we can just avoid more issues in the future instead of getting rid of all of them right now.

Zagrios:
Its ok, as you said is a good first step, I will create an issue especialy to get rid of these issues i think

PR #271, opened 2023-07-01
style: add .prettierignore

RedlineTriad:
Without this it would try to format the release folder. Which would take a *very* long time.

## Scope

This makes it so running prettier doesn't get stuck on the minified javascript in the release folder.

## How to Test

Run `npx prettier -w .` and it doesn't try to format the release folder.

## Notes

With this merged, can you @Zagrios please run the formatter with  `npx prettier -w .` and commit it quickly?
Not doing it myself because it would cause an immense amount of merge conflicts that would need to be fixed if the PR doesn't get merged immediately.

You may have to tweak the prettier configuration to your liking though, as by default it seem to use 2 space indentation.

After this is done I would probably add prettier to CI.

## Emoji Guide

> 🟢 Nice refactor!

| | | |
| --- | --- | --- |
| Blocking | 🔴 ❌ 🚨 | RED |

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+**/dist
+coverage


Zagrios:
i'm not sure i want to use Prettier right now, do you need it for something ?

RedlineTriad:
> i'm not sure i want to use Prettier right now, do you need it for something ?

It's already installed in the project, and without it, contributors can't be sure what correct formatting is supposed to be.

This also causes extra diffs, when everyone uses their own format, and extra work, if the manually have to adjust things to keep diffs clean.

I don't care what formatting is used, but I want it to be consistent, and prettier is also nicely supported by IDEs.

Zagrios (approved):

PR #286, opened 2023-07-16
Chore/rework issue templates

Snailedlt:
Since this is a small PR I'll describe it very briefly instead of following the PR template.

In this PR I simply commented out information that is only useful to the issuer, and not the issue reviewers. This will make it simpler to submit issues, and make reviewing easier since it helps issuers make shorter, more concise issues.

I also changed from bold text to Heading 2 for each subsection in the issue template.

I was unsure whether to keep the full text and comment it out, or simplify it first and comment out. So I did a mix. For the most part the text is the same, but some of it is removed and simplified, so please review it closely :)

### DIFF
--- a/.github/ISSUE_TEMPLATE/-bug--bug-report.md
+++ b/.github/ISSUE_TEMPLATE/-bug--bug-report.md
@@ -6,27 +6,30 @@ labels: bug
 assignees: Zagrios
 ---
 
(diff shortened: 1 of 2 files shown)


Zagrios:
Yeah that better, thanks 😉

Zagrios (approved):

PR #309, opened 2023-08-16
Better spanish translations

GabiRP:
## Scope

Fixed various spanish strings that made no sense in that context.


Zagrios (approved):
great! thanks 👍

PR #325, opened 2023-09-21
Skip Liv initialization on Linux

Insprill:
## Scope

As of 46eb8d0, when launching the application on Linux it would immediately crash due to there not being an implementation of `#isLivInstalled` for Linux.

## Implementation

To fix this, I've simply added a Linux implementation that always returns false.

## How to Test

Try launching the application at/after 46eb8d0 with and without this patch.

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/src/main/services/liv/liv.service.ts
+++ b/src/main/services/liv/liv.service.ts
@@ -25,7 +25,8 @@ export class LivService {
             win32: async () => {
                 const regRes = await regedit.promisified.list([this.livRegeditKey]).then(res => res[this.livRegeditKey]);
                 return regRes?.exists;
-            }
+            },
+            linux: async() => false,
         });
     }
 
--- a/src/main/services/liv/liv.service.ts
+++ b/src/main/services/liv/liv.service.ts
@@ -20,14 +20,13 @@ export class LivService {
 
     }
 
-    public isLivInstalled(): Promise<boolean> {
+    public async isLivInstalled(): Promise<boolean> {
         return execOnOs({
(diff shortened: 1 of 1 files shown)


src/main/services/liv/liv.service.ts
```diff
@@ -25,7 +25,8 @@ export class LivService {
             win32: async () => {
                 const regRes = await regedit.promisified.list([this.livRegeditKey]).then(res => res[this.livRegeditKey]);
                 return regRes?.exists;
-            }
+            },
+            linux: async() => false,
```
Zagrios, line 29: I think it should be better to pass `true` to the `execOnOs` function to not throw error (see function signature)
At least it will handle MacOs as well

indeed I will soon add e2e testing that will run on windows/linux/macos so if e2e tests was already there your PR not pass the checks

Zagrios (changes requested):

src/main/services/liv/liv.service.ts
Insprill: Passing true makes it return undefined which causes an error when calling `then` on the result.

Zagrios: in that case i think you can convert `isLivInstalled` to an async function and it will return a promise in any case

Insprill: Yes, that works. That's an interesting bit of behavior I wasn't aware of, good to know!

Zagrios (approved):
🙌

PR #346, opened 2023-11-07
[feature-274] oculus pcvr download

Zagrios:


lorke1:
Is it not yet finished?
If it is finished how do I use it?

Zagrios:
Hi, We will release this, in an alpha most likely this weekend

PR #369, opened 2023-12-04
[bugfix] Don't download map already installed when installing a playlist

Zagrios:


Zagrios:
Was merged bug GitHub tell me its not merged ¯\_(ツ)_/¯

PR #417, opened 2024-02-04
[bugfix-416] Wait for SteamVR folder rename before exit

### LINKED ISSUE #416: [BUG] : Launching FPFC via shortcut fails to restore SteamVR.
## Bug Description
When launching Beat Saber through a bs-manager shortcut in FPFC, it will rename the SteamVR folder to SteamVR.bak. Most of the time it will fail to rename it back. Failure seems random but looks like a race condition with bs-manager exiting before waiting for rename.

## Reproduction Steps
1. Create a launch shortcut for any Beat Saber version.
2. Close bs-manager if already running.
3. Launch via shortcut.
4. ~3/4 times on my machine the SteamVR folder will remain as SteamVR.bak.

## System Specs
- OS: Windows 10 version 22H2
- bs-manager: v1.4.4

slinkstr:
## Scope

Adds await to restoreSteamVr call to prevent program exit before rename completes.

[Closes #416](https://github.com/Zagrios/bs-manager/issues/416)


slinkstr:
Realized I blundered when resolving sonarcloud warnings... Fixed.

Zagrios (approved):
Seems good to me 🙌

PR #439, opened 2024-03-17
Fix BSIPA not running on Linux

### LINKED ISSUE #440: [BUG] : Mods cannot be installed on Linux
## Bug Description
Mods cannot be installed on Linux as BSIPA fails to run.

## Reproduction Steps
- Try installing mods on Linux.

## Expected Behavior
Mods should be able to be installed on Linux.

## System Specs
-  OS: Arch Linux
-  Version: v1.4.7

Insprill:
## Scope

Fixes the inability to install mods under Linux due to BSIPA being executed with Windows-specific parameters.
Closes #440.

## Implementation

When running on Linux, BSIPA is now run with dotnet in a screen session. The screen session is a hack. However, it's required since BSIPA tries to move the console cursor, which doesn't exist in our environment, and crashes with an ArgumentOutOfRangeException before exiting. Running it in a screen session fixes this problem.

## How to Test

- Try installing mods on Linux

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/src/main/services/mods/bs-mods-manager.service.ts
+++ b/src/main/services/mods/bs-mods-manager.service.ts
@@ -155,8 +155,13 @@ export class BsModsManagerService {
         }
 
         return new Promise<boolean>(resolve => {
-            log.info("START IPA PROCESS", `start /wait /min "" "${ipaPath}" ${args.join(" ")}`);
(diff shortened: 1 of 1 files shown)


Zagrios:
Thanks 🙌 Will be deployed in the next release 😉

Zagrios (approved):

PR #449, opened 2024-03-26
Fix Steam elevated check on Linux

Insprill:
I see the v1.5.0 branch is what's currently active so I've targeted this PR to that. Let me know if you'd like it on master and I'll rebase.

## Scope

This removes the forced skip of the elevated check for Steam on Linux.
Steam should never be running as admin on Linux, so defaulting to true doesn't make sense, and the existing isElevated check appears to be working correctly.

## Emoji Guide

> 🟢 Nice refactor!


Zagrios (approved):

Zagrios:
Thanks! 🙌

PR #451, opened 2024-03-27
Add support for launching Beat Saber on Linux

### LINKED ISSUE #441: [BUG] : Beat Saber cannot be launched on Linux
## Bug Description
Beat Saber cannot be launched on Linux.

## Reproduction Steps
Try to launch Beat Saber on Linux.

## Expected Behavior
Beat Saber should launch on Linux using a user-selected (or Steam default?) version of Proton.

## System Specs
-  OS: Arch Linux
-  Version: v1.4.7

Insprill:
## Scope
This PR adds support for launching Beat Saber on Linux.

Closes #441.

## Implementation
A new application setting has been added, "Proton Path", which the user must specify before the game can be launched. This setting will only be shown to Linux users. When the game is launched, instead of running the Beat Saber executable directly, Proton will be run with the appropriate environment variables and arguments.

This is quite a primitive implementation—it leaves out many environment variables that Steam normally passes to the game (mostly related to shader caching) and doesn't allow the user to pass on their own, which is commonly done and sometimes necessary to get Proton games to run on certain systems. This should, however, work fine for most users. In my testing on Arch, everything works flawlessly.

Running as admin is not currently supported.

## How to Test (Linux)
*Note: Testing on Windows should also be done to ensure this PR doesn't cause any regressions.*
- Set the Proton path in settings.
- Launch the game.

## Emoji Guide

> 🟢 Nice refactor!


Zagrios:
Thanks a lot ! 🎉🔥 Tested on windows and not problem 👌 
@Insprill do you think that BSM is stable enough now on Linux to ship an AppImage? (not in production first of course, it will be shipped with the next alpha 😅)

Insprill:
From my testing, yes! Everything is working fine on my system. I'll be happy to continue maintaining Linux support if (probably when :sweat_smile:) any issues pop up as well :)

PR #464, opened 2024-04-20
[chore] use winhttp.dll to detect bsipa installation

Zagrios:


src/main/services/mods/bs-mods-manager.service.ts
```diff
@@ -112,7 +112,7 @@ export class BsModsManagerService {
 
     private async getBsipaInstalled(version: BSVersion): Promise<Mod> {
         const bsPath = await this.bsLocalService.getVersionPath(version);
-        const injectorPath = path.join(bsPath, "Beat Saber_Data", "Managed", "IPA.Injector.dll");
+        const injectorPath = path.join(bsPath, "IPA", "winhttp.dll");
```
Meivyn, line 115: That technically wouldn't guarantee that BSIPA is installed. You're better looking for that file in the root folder directly.

Zagrios: Hi, I did that after a suggestion to have a bit more parity with MA, which uses winhttp.dll to detect if BSIPA is installed ([reference](https://github.com/bsmg/ModAssistant/blob/2eae530926f6f373805fdfcb665d2ed9e8aa45eb/ModAssistant/Pages/Mods.xaml.cs#L402)). 
I beleave MA has more experience with anything related to mods, so I think it's not a bad idea 🤔

Meivyn: MA does use the injector to know whether it's installed or not (as you can see [here](https://github.com/bsmg/ModAssistant/blob/2eae530926f6f373805fdfcb665d2ed9e8aa45eb/ModAssistant/Pages/Mods.xaml.cs#L250)).

That line you linked is only hit when the user presses the install button. It's meant to let BSIPA update itself on the next game launch and prevent MA from force-installing it if it's already installed. Also, note that it doesn't make use of the `IPA` folder.

Zagrios: Oh you're right I've missed that 👀 
I think i will revert this commit then 👌

PR #477, opened 2024-05-14
GitHub Actions Build Workflow

eliasbenb:
## Scope

I wanted to test a feature that wasn't released yet but also didn't want to setup the dev environment on my personal computer. So, I wrote a simple GH actions build workflow to build and package the electron executable for me.

Since GH Actions is free, I thought this would be a nice addition for future BSManager testers.

## Implementation

I made the build workflow automatically build all commits on the master branch. Ideally this can be set to all other branches too. After the electron package is built, it is stored inside a GitHub artifact. Note that these expire after 90 days.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,34 @@
+# This workflow will...
+
+name: Build
+
+on:
+    push:
+        branches: ["master"]
+    workflow_dispatch:
+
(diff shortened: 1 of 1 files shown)


Zagrios (approved):

Zagrios:
Yeah that's useful, thanks 🙌

PR #509, opened 2024-07-13
Fix Linux build & crash on startup

Insprill:
## Scope

Fixes builds failing on Linux, and it crashing on startup

## Implementation

The same as https://github.com/Zagrios/bs-manager/pull/498

I found this NPM issue, which seems to be what we're running into here, although deleting the `node_modules` folder doesn't fix the issue for me - only manually adding it to the lock file does.
https://github.com/npm/cli/issues/4828

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11201,6 +11201,23 @@
                 "esbuild": ">=0.12 <1"
             }
         },
+        "node_modules/esbuild/node_modules/@esbuild/linux-x64": {
+            "version": "0.20.1",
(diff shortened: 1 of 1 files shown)


Zagrios:
@Insprill are you sure this fix is needed 🤔
The github action is building for linux successfully each time 🤔
https://github.com/Zagrios/bs-manager/actions/runs/9928893690

Zagrios (approved):

Issue #546, opened 2024-08-03
[BUG] : regedit-rs breaks Linux Support again


silentrald:
## Bug Description
Similar to issue #411, can't run the current version of bs-manager thru node in linux. Attached the stack error logs below.

```
[1] [electronmon] waiting for a change to restart it
[1] App threw an error during load
[1] Error: Cannot find module 'regedit-rs-linux-x64-gnu'
[1] Require stack:
[1] - <file path>/bs-manager/release/app/node_modules/regedit-rs/js-binding.js
[1] - <file path>bs-manager/release/app/node_modules/regedit-rs/index.js
[1] - <file path>/bs-manager/.erb/dll/main.bundle.dev.js
[1]     at Module._resolveFilename (node:internal/modules/cjs/loader:1152:15)
[1]     at s._resolveFilename (node:electron/js2c/browser_init:2:121098)
[1]     at Module._load (<file path>/bs-manager/node_modules/runtime-required/runtime-required.js:28:44)
[1]     at Module.require (node:internal/modules/cjs/loader:1240:19)
[1]     at require (node:internal/modules/helpers:179:18)
[1]     at Object.<anonymous> (<file path>/bs-manager/release/app/node_modules/regedit-rs/js-binding.js:1:3504)
[1]     at Module._compile (node:internal/modules/cjs/loader:1373:14)
[1]     at Module._extensions..js (node:internal/modules/cjs/loader:1432:10)
[1]     at Module.load (node:internal/modules/cjs/loader:1215:32)
[1]     at Module._load (node:internal/modules/cjs/loader:1031:12)
[1] bs-manager encountered an error
[1] Error: Cannot find module 'regedit-rs-linux-x64-gnu'
[1] Require stack:
[1] - <file path>/bs-manager/release/app/node_modules/regedit-rs/js-binding.js
[1] - <file path>/bs-manager/release/app/node_modules/regedit-rs/index.js
[1] - <file path>/bs-manager/.erb/dll/main.bundle.dev.js
[1]     at Module._resolveFilename (node:internal/modules/cjs/loader:1152:15)
[1]     at s._resolveFilename (node:electron/js2c/browser_init:2:121098)
[1]     at Module._load (<file path>/bs-manager/node_modules/runtime-required/runtime-required.js:28:44)
[1]     at Module.require (node:internal/modules/cjs/loader:1240:19)
[1]     at require (node:internal/modules/helpers:179:18)
[...]

PR #547, opened 2024-08-03
[bugfix] exclude regedit-rs on externals when webpack bundles it in linux

silentrald:
## Scope

[closes TICKET-546](https://github.com/Zagrios/bs-manager/issues/546)

## Implementation

Issue was that when **webpack** bundles the project and **[name].bundle.dev.js** was executed, `require('regedit-rs')` was being called, producing the error in the ticket.

image

Resolution was to exclude `regedit-rs` within the webpack config `external` property.

## Emoji Guide

> 🟢 Nice refactor!


Zagrios:
Tested, thats work great, thanks a lot! 🙌

Zagrios (approved):

Issue #560, opened 2024-08-14
[FEAT.] : Ask users to select their installation folder on startup


silentrald:
## Problem
When a user manually moves the installation folder to another folder/disk, bs-manager still uses the default or installation-folder property in config.json.

## Solution
If the installation folder (BSManager) does not exist on startup, prompt the user a modal to select either a new or existing installation folder. Default installation folder value should still be used on the prompt.

## Workaround
* Manually editing the config.json to replace the installation-folder value to the new location.
* Change the installation folder within BSManager in the settings
  * **NOTE:** Does not work if the original folder location does not exist anymore, since BSManager wants to transfer the original installation folder to the new location.

## Additional Context
Issue reported in https://discordapp.com/channels/1049624409276694588/1273029459544375459

Might be related to #481

Issue #481, opened 2024-05-19
[BUG] : Not possible to change Installation Folder


akama-aka:
## Bug Description
Hello, my SSD with my BS Manager Files use normally "F:" but somehow Windows Changed it once to "V:". Now I had to change my SSD back to the "F:" letter because my Maps are Physical Symblinked to the BeatSaber Data Directory to prevent Data Loss and other issues. But as soon as I tried to change the BS Manager Installation Folder back to the old one I get everytime the error that it wasnt successfull. The reason is that it cant find the "V:" Partition. Is it possible to add a checkmark to disable the BS Manager Files relocation?

## Reproduction Steps
1. Use the one SSD/HDD you have rn for your BS Manager Installation.
2. Change the Partition Letter to a different in the Windows Partition Manager
3. Reboot or go back to BS Manager and try to change the Installation Location.

## Expected Behavior
It should change the directory and load the Versions

## Screenshots
_Not really needed_

## System Specs
-   OS: Windows 
-   Version: 22H2 19045.4412

## Additional context
Logs:

```logs
[2024-05-19 01:56:04.040] [error] Error: ENOENT: no such file or directory, scandir 'V:\BSManager' ENOENT
[2024-05-19 01:57:59.417] [info]  [OUTPUT] isLivInstalled() false
[2024-05-19 01:57:59.433] [error] Unhandled Exception UnhandledRejection Error: ENOENT: no such file or directory, mkdir 'V:\BSManager'
    at mkdirSync (node:fs:1398:3)
    at Object.mkdirSync (node:electron/js2c/asar_bundle:2:12611)
    at e.exports._ensureDirectory (C:\Users\...\AppData\Local\Programs\bs-manager\resources\app.asar\dist\main\main.js:8:137978)
    at e.exports.get store [as store] (C:\Users\...\AppData\Local\Programs\bs-manager\resources\app.asar\dist\main\main.js:8:136719)
    at new N (C:\Users\...\AppData\Local\Programs\bs-manager\resources\app.asar\dist\main\main.js:8:134359)
    at new e.exports (C:\Users\...\AppData\Local\Programs\bs-manager\resources\app.asar\dist\main\main.js:14:72899)
    at o.initStore (C:\Users\...\AppData\Local\Programs\bs-manager\resources\app.asar\dist\main\main.js:14:1032488)
[2024-05-19 01:58:00.562] [info]  Checking for update
[...]

PR #566, opened 2024-08-23
[feat-560] ask user for installation folder/path

silentrald:
## Scope

[closes TICKET-560](https://github.com/Zagrios/bs-manager/issues/560)
[closes TICKET-481](https://github.com/Zagrios/bs-manager/issues/481)

## Implementation

Added a way to ask the user for an existing or non-existing installation folder/path

Modified `modal.component.tsx` to add the property **closable**. If **closable = false**, then you can't exit out the modal by any means, except by adding a button to resolve the modal itself.

Created `AskInstallPathModal` (ask-install-path-modal.component.tsx).

Created `SetupService` (setup.service.ts) class to handle any user configuration before application startup. Logic such as sequential modals/prompts can be added here sequentially to ask for config values. Not sure if one whole modal or separate modals will be done but should be easy enough to edit the logic there.

Created `bs-installer-ipcs.ts` to handle all the server calls for the installation folder/path. Transferred some codes from `bs-download-ipcs.ts` to here.

Might need to transfer the the installation folder/path code from `SteamDownloaderService` to another class as well, but might do it in another PR.

## Screenshots

image

*UI may still change depending on suggestions*
As of commit 9e2f716c53d61eccc8d9e247de4af8d415887567

### Test Case

1. Edit the **installation-folder** property in the `config.json` file with any existing or non-existing folder path other than your original installation folder.
- Windows: `%AppData%/Roaming/bs-manager`
- Linux: `/.config/bs-manager`
2. Open BSManager and a modal will appear (similar to the above screenshot)
3. Select any folder and click on **Confirm**
4. On close either:
4.1. If existing installation folder: version should be shown on the sidepanel
4.2. If non-existing folder: installation folder will be created

### Additional Tests
- Check if the **Default** button works properly.
- Check if this would not ask the user if the installation path points to an existing installation folder.
- Check if the songDetailsCache was loaded properly.

## New/Updated Resource Text

- [x] **misc.confirm:** Confirm
- [x] **pages.settings.installation-folder.description:** Change the folder that will contain all the content downloaded by BSManager.
- [x] **modals.ask-install-path.title:** Installation folder
- [x] **modals.ask-install-path.choose-folder-description:** Choose the folder that will contain all the content downloaded by BSManager.
- [x] **modals.ask-install-path.choose-folder:** Choose folder
- [x] **modals.ask-install-path.default:** Default
- [x] **modals.ask-install-path.default-tooltip:** Defaults to your home folder

**NOTE:** ticked are already translated to other languages.

## Emoji Guide

> 🟢 Nice refactor!


Zagrios:
@silentrald Maybe its a good its to add an explaining text between the title and the "choose folder" field 🤔 
Something like "Choose the folder that will contain all the content downloaded by BSMananger" (its juste an exemple but yeah something like that)

silentrald:
@Zagrios that should be fine, a little bit of adjustment would be "Choose the folder that will contain all the contents downloaded by BSManager".

Might need to update the settings page explanation text as well if that's the case, to pattern these two. Will still use the above text and change "Choose" to "Change". Although, what do you think, should I change it or leave it as is?
image

Also for the translation, should I ping someone in the discord channel so that I can fit it in the PR, or do we delay the translation and let someone just translate it before the next alpha release?

Zagrios:
@silentrald you can change it, it's sounds good to me 👍 

I also notice that you made change to the cache system to wait for the user choosing a folder, thats made me wondering if I should move those cache files to the installation folder of BSM that's will always be on the "main" drive of the user 🤔 Cause all the cache files are there to improve perfomances, and the user's "main" drive is more likely to be an hight performance drive (like ssd rather than a hdd or a local network drive) 🤔

Zagrios:
@silentrald Also maybe add a Tooltip on on the "Default" button to indicate that this will chose the home folder of the user
(you can find a lot of exemples searching for `Tippy` in the code; You will probably need to add a div wrapper for the button but I'll let you see by yourself)

silentrald:
> silentrald you can change it, it's sounds good to me 👍
> 
> I also notice that you made change to the cache system to wait for the user choosing a folder, thats made me wondering if I should move those cache files to the installation folder of BSM that's will always be on the "main" drive of the user 🤔 Cause all the cache files are there to improve perfomances, and the user's "main" drive is more likely to be an hight performance drive (like ssd rather than a hdd or a local network drive) 🤔

Yeah I think moving it to the installation folder might be better. Something like a `.cache` or `Cache` folder might suffice but might need be handled in another PR since this is out of scope of the current feature.

Also context for why I did that as well, had to delay the instantiation of the ElectronStore object since this automatically creates the installation folder and the `config.cfg` file, which makes the modal asking for the user for an installation folder not show up. This makes the configuration service not yet readable/writable until setup is done, delaying the loading the cache files. Although this doesn't feel like a bug since there is a chance to read the config file on the installation folder that the user selected.

src/shared/models/ipc/ipc-routes.ts
```diff
@@ -43,6 +41,12 @@ export interface IpcChannelMapping {
     "bsv-search-playlist": {request: PlaylistSearchParams, response: BsvPlaylist[]};
     "bsv-get-playlist-details-by-id": {request: {id: string, page: number}, response: BsvPlaylistPage};
 
+    /* ** bs-installer-ipcs ** */
+    "bs-installer.folder-exists": { require: void, response: boolean };
```
Zagrios, line 45: You wrote `require` here
```suggestion
    "bs-installer.folder-exists": { request: void, response: boolean };
```

src/renderer/components/shared/bsm-button.component.tsx
```diff
@@ -84,18 +93,35 @@ export const BsmButton = forwardRef<unknown, Props>(({ className, style, iconSty
 
     const handleClick = (e: MouseEvent<HTMLDivElement>) => !disabled && onClick?.(e);
 
+    const renderTooltip = () => {
+        return (
+            <Tippy
+                className="!bg-main-color-1"
+                content={t(tooltip)}
```
Zagrios, line 100: I meant, adding a tooltip while hovering the button globally, like this users are sure to see the tooltip.
So no need to add all this stuff in the button component

src/main/ipcs/setup.ipcs.ts
```diff
@@ -0,0 +1,11 @@
+import { from } from "rxjs";
+
+import { IpcService } from "main/services/ipc.service";
+import { SongDetailsCacheService } from "main/services/additional-content/maps/song-details-cache.service";
+
+const ipc = IpcService.getInstance();
+
+ipc.on("setup-done", (_, reply) => {
```
Zagrios, line 8: Why do we need to load the `songs details cache` here ? since it do not rely on the installation path ?

src/renderer/services/setup.service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    private async checkInstallationPath(): Promise<void> {
+        try {
+            const exists = await lastValueFrom(this.ipcService.sendV2("bs-installer.folder-exists"))
+            if (exists) {
+                return;
+            }
+
+            let modalResponse: ModalResponse<{ installPath: string }> = { exitCode: ModalExitCode.NO_CHOICE };
+            while (modalResponse.exitCode !== ModalExitCode.COMPLETED) {
```
Zagrios, line 53: Do we still need the while loop here ? since the modale is marked as non closable ?

Zagrios (changes requested):

src/renderer/components/shared/bsm-button.component.tsx
silentrald: Ahhh, sorry about that, thought we needed an icon for it, will apply a commit for this.

src/main/ipcs/setup.ipcs.ts
silentrald: `SongDetailsCacheService` requires `ConfigurationService`. The configuration service needs the `config.cfg` file from the installation path in order for it to read the **etag** value.

image

src/renderer/services/setup.service.ts
silentrald: Yeah, I think the loop there can be removed as well

src/main/ipcs/setup.ipcs.ts
Zagrios: Oh yeah you're right i was forgetting the etagKey 👌

Zagrios:
@silentrald can you resove confict ?
No need for the cache to wait the setup, we have now a "staticConfig"

silentrald:
@Zagrios rebased and resolved the conflicts. Also removed the waiting on setup as well.

Zagrios:
@silentrald I've found a little bug, when I select a folder from the modal, the installation path doesn't change in the settings, and I have this error:

```text
Error: ENOENT: no such file or directory, scandir 'D:\\BSManager' at async SetupService.check (http://localhost:1212/renderer.dev.js:248304:13)
```

See gif:

little bug

silentrald:
@Zagrios yep, seems that I haven't covered the case where the oldDir doesn't exist in the `InstallationLocationService.setInstallationDirectory`. Tested in my windows VM, can set the new folder without moving a non-existent one.

Zagrios (approved):
That seems good to me 🙌 
Thanks !

PR #572, opened 2024-09-01
[feature] add advanced settings

Zagrios:
- Add advanced setting to disable Hardware Acceleration
- Add advanced setting to use Symlinks instead of junctions


FelixKainz:
Stupid question: why would I want to disable HW acceleration? Is it about performance of BS or stability?

PR #575, opened 2024-09-05
[feat-75] support drag and drop for map zip files

### LINKED ISSUE #75: [FEAT.] : Drag drop map to install it
**Is your feature request related to a problem? Please describe.**
We can't manually install maps on a version.

**Describe the solution you'd like**
Drag and drop folder or zip of a map onto the liste of maps to install it

**Describe alternatives you've considered**
No alternative

**Additional context**

drag drop map

silentrald:
## Scope

closes #75

## Implementation

Implemented drag and drop for map zip files.

Added `Dropzone` component to support drag and drop of files.

Created `processZip` in zip helper file to safely handle zip files, although SonarCloud doesn't seem to like the handling.

Added `LocalMapsManagerService.importMap` to import the zip file and copied almost the same code in downloadMap for pathing and caching.

## Screenshots

image

## How to Test

**Drag and drop the following files:**

* Valid map zip file
* Invalid zip file
* Any non-zip file

**Check if it will be installed in either the specific version folder or shared folder as well.**

**Check if switching to the playlist tab is not affected by the dropzone.**


src/main/services/additional-content/maps/local-maps-manager.service.ts
```diff
@@ -306,6 +309,85 @@ export class LocalMapsManagerService {
         return null;
     }
 
+    public importMaps(zipPaths: string[], version?: BSVersion): Observable<Progression> {
+        const progress: Progression<string> = {
```
Zagrios, line 313: This function should handle zips containing multiple maps, not just one zip per map.\
It should be compatible with the export maps feature of BSM, which creates a single zip file for all the selected maps.

src/main/services/additional-content/maps/local-maps-manager.service.ts
```diff
@@ -306,6 +309,85 @@ export class LocalMapsManagerService {
...
+                    } catch (error: any) {
+                        log.error(`Could not import "${zipPath}"`, error);
+                        this.ipc.send<{map: string, error: string}>(
+                            "map-not-imported",
+                            this.windows.getWindows("index.html").at(0),
+                            { map: zipPath, error: error.code }
+                        );
+                    }
+                }
+                observer.complete();
```
Zagrios, line 337: Prefer placing the .complete() in the finally block and add an .error() in the catch of the promise.\
This way, it will handle all potential errors that are not caught by your try/catch block.\
(Even though, I agree, in the current state, unhandled errors are shouldn't possible)

src/main/services/additional-content/maps/local-maps-manager.service.ts
```diff
@@ -306,6 +309,85 @@ export class LocalMapsManagerService {
...
+                for (const zipPath of zipPaths) {
+                    ++progress.current;
+                    progress.data = zipPath;
+                    observer.next(progress);
+
+                    try {
+                        await this.importMap(zipPath, version);
+                    } catch (error: any) {
+                        log.error(`Could not import "${zipPath}"`, error);
+                        this.ipc.send<{map: string, error: string}>(
```
Zagrios, line 330: Its preferable to remove this `ipc.send` cause' it can spam the user if all dropped zips are invalid.\
Just handle errors on the frontend as normally and show a warning if the progression does not reach 100%
(This is even truer when this function will handle zips of multiple maps)

src/renderer/services/maps-manager.service.ts
```diff
@@ -37,13 +37,33 @@ export class MapsManagerService {
...
+
     private constructor() {
         this.ipcService = IpcService.getInstance();
         this.modal = ModalService.getInstance();
         this.progressBar = ProgressBarService.getInstance();
         this.notifications = NotificationService.getInstance();
         this.config = ConfigurationService.getInstance();
         this.linker = VersionFolderLinkerService.getInstance();
+
+        this.ipcService.watch<{map: BsmLocalMap, version?: BSVersion}>("map-imported")
```
Zagrios, line 51: You don't need these watchers.\
I will deprecate the watch function in the future.\
Using it too much overcomplicates the code; the backend should never send random events like this. It should only send events when a function is called, until the end of the function’s execution.

In your `importMaps`, the `"bs-maps.import-maps"` should return a `Progression<BsmLocalMap>`, and then you can do something like this:
```ts
sendV2("bs-maps.import-maps", {...args}).pipe(
  tap( progress => importListerners.foreach(listener => listener(version, progress.data))
)
```

src/renderer/services/maps-manager.service.ts
```diff
@@ -37,13 +37,33 @@ export class MapsManagerService {
...
+            .subscribe(data => {
+                ++this.importCount;
+                this.importListeners.forEach(listener => listener(data.map, data.version));
+            });
+
+        // NOTE: Generic notify call from server side
+        this.ipcService.watch<{map: string, error: string}>("map-not-imported")
```
Zagrios, line 58: Same here, but this should be removed anyway due to the `spam of notifications`

src/renderer/services/maps-manager.service.ts
```diff
@@ -162,6 +213,17 @@ export class MapsManagerService {
         return lastValueFrom(this.ipcService.sendV2("unregister-maps-deep-link"));
     }
 
+    public addImportListener(listener: (map: BsmLocalMap, version?: BSVersion) => void): void {
+        this.importListeners.push(listener);
+    }
+
+    public removeImportListener(listener: (map: BsmLocalMap, version?: BSVersion) => void): void {
+        const index = this.importListeners.indexOf(listener);
```
Zagrios, line 221: I wonder what happen if the same callback is added multiple time 🤔 \

Like:
```ts
const cb = () => {};
addImportListener(cb);
addImportListener(cb);
addImportListener(cb);

removeImportListener(cb);
```
Maybe a `Set` of callbacks would work better here instead of an array 🤔
I let you test that 😉

Zagrios (changes requested):

src/main/services/additional-content/maps/local-maps-manager.service.ts
silentrald: Is there a way to currently verify that the exportMaps zip file is generated by BSM? Something like a `metadata` within the zip file.

Also the only problem that I can see here is the **zip bomb check** I've implemented in processZip. Where there is a limit to the file count and file size. We can add that information in within the metadata I guess to bypass the check and a warn users about downloading other users exportMaps zip files.

src/renderer/services/maps-manager.service.ts
silentrald: Yeah might be better to do a `Set` in any case, but we'll see if that works.

---

Actually there was a bug is this code as well, where the listeners weren't removed properly. Not sure if the fix should be here since the PR might take some more time to get merged, but the snippet bugged code is here.

In master currently
`local-maps-list-panel.component.tsx`
```js
    useEffect(() => {
        if (isActiveOnce) {
            loadMaps();
        }

        return () => {
            setMaps(null);
            loadPercent$.next(0);
            subs.forEach(s => s.unsubscribe());
            mapsDownloader.removeOnMapDownloadedListener(loadMaps); // should not be loadMaps
        };
    }, [isActiveOnce, version, linked]);

    useEffect(() => {
        if(isActiveOnce){
            // The listener that was being added multiple times
            mapsDownloader.addOnMapDownloadedListener((map, targetVersion) => {
                if (!equal(targetVersion, version)) {
                    return;
                }
                setMaps((maps ? [map, ...maps] : [map]));
            });
        }

        return () => {
            mapsDownloader.removeOnMapDownloadedListener(loadMaps); // should not be loadMaps
        }

    }, [isActiveOnce, version, maps])
```

src/renderer/services/maps-manager.service.ts
silentrald: Got this, might just do this on `import-maps` and not on `download-maps`. I can do the refactoring on download maps once I get this PR merged.

src/renderer/services/maps-manager.service.ts
Zagrios: 👀 Yeah I will fix that in on master 😅

src/renderer/services/maps-manager.service.ts
Zagrios: Let me this 👌, there is few more places that use this function, I will try to remove them all today

src/main/services/additional-content/maps/local-maps-manager.service.ts
Zagrios: We can add a metadata file to the generated zip, but users can totally create zips that contain multiple maps on their own for plenty of reasons.

We can maybe split the zip verification to another ipc, and warn the user with a modal if the zip exeed a certain size or a certain amount of maps 🤔 

Or we could let the user be responsible for what kind of zips they are importing (I mean, if they want to import 1TB of maps, they should be able to do it)

silentrald: > We can add a metadata file to the generated zip, but users can totally create zips that contain multiple maps on their own for plenty of reasons.

Yeah fair point, the metadata would be either useless or redundant in any case. Might just do a simple verification on each folder on depth 1 if they contain the "I/info.dat" file.

> We can maybe split the zip verification to another ipc, and warn the user with a modal if the zip exeed a certain size or a certain amount of maps 🤔
Or we could let the user be responsible for what kind of zips they are importing (I mean, if they want to import 1TB of maps, they should be able to do it)

I think the latter is probably the better option since the user should be responsible for what the zip files that are being imported is.

I've implemented that check is because sonarcloud suggested it so might ignore that in the future in any case.

Zagrios: Yeah, I think SonarCloud's suggestion is great when it’s for a website, but it doesn’t really apply to BSM

silentrald:
@Zagrios Side tangent, seems that the call from `observer.error` shows the call stack error in the renderer. Seems the same as well for the other sendV2 <-> Observer interaction in the codebase as well. Might be a bug in the integration? But might check with you if this is expected.

image

image

Zagrios:
> @Zagrios Side tangent, seems that the call from `observer.error` shows the call stack error in the renderer. Seems the same as well for the other sendV2 <-> Observer interaction in the codebase as well. Might be a bug in the integration? But might check with you if this is expected.
> 
> image
> 
> image

Yeah you need to handle the error in the frontend as well

silentrald:
> Yeah you need to handle the error in the frontend as well

Already did in the latest commit, the notification popup shows in the background when I close the error stack in the renderer so I'm not sure what part I've missed 🤔.

https://github.com/Zagrios/bs-manager/blob/1ccaad2e18d9f7c93c7830ced5350248aeed868e/src/renderer/services/maps-manager.service.ts#L163-L215

Zagrios:
> > Yeah you need to handle the error in the frontend as well
> 
> Already did in the latest commit, the notification popup shows in the background when I close the error stack in the renderer so I'm not sure what part I've missed 🤔.
> 
> https://github.com/Zagrios/bs-manager/blob/1ccaad2e18d9f7c93c7830ced5350248aeed868e/src/renderer/services/maps-manager.service.ts#L163-L215

The error comes from this line `this.progressBar.show(importProgress$, true);` Since the progressBarService shares the same observer, it also receives the error from it.
You need to derive the main observer to create another one that doesn't receive any errors.

You can check the `handleDownload()` function in `oculus-downloader.service.ts`

silentrald:
Ahhh gotcha, will amend the last commit so that the PR should be ready for a re-review.

Zagrios:
@silentrald I've just pushed some fixes, but I have to go.\
I'm still encountering an error when I drop my zip file, as it doesn't find any Info.dat (see the gif).
zip dont work
I get this error message:
```
[1] 17:37:30.357 > Zip file "C:\Users\Mathieu\Desktop\1.37.3Maps.zip" does not contain any "Info.dat" file
[1] 17:37:30.358 > Error: No "Info.dat" file located in any of the zip files
```
The `const files = zip.file(this.INFO_DAT_REGEX);` results as empty array.

silentrald:
@Zagrios might be related to the path separator. Not sure if Windows uses `/` or `\`, couldn't check this on my linux machine, though let me try to fix this in a Windows VM.

silentrald:
@Zagrios got it to work on Windows.

Lost your commit though when I did a full rebase but I tried to reapply it on commit [615ac04](https://github.com/Zagrios/bs-manager/pull/575/commits/615ac0496a231eca5c53ec4abe153e2ebd964b8a). Might need to double check that, if I missed some lines, you can cherry pick the commit on your local to this PR.

src/main/services/additional-content/maps/local-maps-manager.service.ts
```diff
@@ -334,25 +334,30 @@ export class LocalMapsManagerService {
...
                     if(error) {
                         log.error("Could not read zip entries", zipPath, error);
-                        zip.close();
+                        await zip.close();
                         continue;
                     }
 
                     const mapsFolders = zipEntriesValues.reduce((acc, entry) => {
-                        if(!/(^|\/)(I|i)nfo.dat$/.test(entry.name)){ return acc; }
+                        if(!/(^|\/)[I-i]nfo\.dat$/.test(entry.name)){ return acc; }
```
silentrald, line 360: Shouldn't this be `(I|i)` not `[I-i]` since the latter means that any character between I and i, so `Jnfo.dat` or `Knfo.dat` would pass this test.

Zagrios: Yep, you'r right 😅 I've been fooled by sonarcloud x)
I will remove the `-` 👌

silentrald: Ohh yeah that's much simpler

Zagrios:
@silentrald can you test to confirm if its still working on Linux ?

silentrald:
@Zagrios can confirm, works on Linux 👌

Zagrios (approved):

Zagrios:
Merging 🔥🔥

PR #584, opened 2024-09-21
[bugfix] handle undefined modal options properly

silentrald:
## Scope

If the modal options is not defined the ff. error is reported. Just handled it in the code so that the check is done properly.

```
[2024-09-21 12:06:03.298] [error] [
  [
    "Uncaught TypeError: Cannot read properties of undefined (reading 'closable')",
    'file:///C:/Users/<user>/AppData/Local/Programs/bs-manager/resources/app.asar/dist/renderer/renderer.js',
    2,
    2311860,
    "Error: Cannot read properties of undefined (reading 'closable')"
  ]
]
```

## How to Test

Open the modal for downloading a BS Version.

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/src/renderer/components/modal/modal.component.tsx
+++ b/src/renderer/components/modal/modal.component.tsx
@@ -17,7 +17,7 @@ export function Modal() {
 
     useEffect(() => {
         const onEscape = (e: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


Zagrios (approved):
Nice 👍

Issue #578, opened 2024-09-11
[FEAT.] : make linux AppImage work out of the box


silentrald:
## Problem

Currently with the linux AppImage, to install mods properly in linux distros, there are hidden dependencies that aren't very well documented. Users need to install both **screen** and **dotnet-3.1.426** version. **screen** package is very trivial for most distros however the specific version of **dotnet** is very hard for most users to get installed.

These are some of the **dotnet** installation process:
* [Microsoft Docs](https://learn.microsoft.com/en-us/dotnet/core/install/linux)
* [Arch Wiki](https://wiki.archlinux.org/title/.NET#Install_multiple_versions_manually)

## Solution

Add the linux executable binaries for **screen** and **dotnet** to the `assets/scripts` folder to properly run the `IPA.exe` executable. This is also feasible since I've got a working branch for this. Only downside here is that the **dotnet** binary is very big, around ~200mb. Will try to update this to get the final AppImage size if I include these changes.

## Alternative solutions (if any)

* For Arch users, I think Insprill created an AUR package which uses **yay** package manager. The dependencies are linked here so it would work properly.

* **wine** can also be an alternative to run `IPA.exe` although **wine** can fail in execution and/or log out error messages unlike **dotnet**.

* _[Unlikely]_ A possibility of BSIPA being compiled as a linux binary so that we wouldn't depend on **dotnet**, minimum install would only be **screen**. 

## Additional context

Feature suggestion is because of Steam Deck users since its very hard to install packages in the handheld console and most users of these are not technically adept to install these stuff.

Zagrios:
Can't we get rid of the screen dependency ? 🤔 
I know we need it because BSIPA tries to move the cursor, but I also had this problem on Windows because I absolutely wanted the app to run without a window... Now it opens a window when executing BSIPA on Windows and no problem

Can't we execute just a command like `dotnet bsipa.exe` in a way that it opens its own window and does not longer throw the cursor moving error?🤔

silentrald:
I think the `screen` and `dotnet` combination might be harder to debug tbh since `screen` silently proceeds execution even if `dotnet` fails. Can't seem to make `dotnet bsipa.exe` to work in any case. Although, I got `mono` or `wine` to work and remove the cursor moving error. I've pasted the logs for each so that we can have a comparison.

<details>
<summary><b>mono logs</b></summary>

```
[1] 09:15:52.745 › executeBSIPA 1.29.1 [ '-n' ]
[...]

Issue #585, opened 2024-09-22
[BUG] : Linux doesn't show unicode titled maps


silentrald:
## Bug Description

When downloading maps with unicode characters, they do not show up in the song list within Beat Saber.

## Reproduction Steps

Download any song with BSManager with any unicode character. Sample map to download:

* みにまむ - misterlihao
* ツユ - アンダーヒロイン - wangyufy

## Screenshots

Error logs from Beat Saber
![366446147-7c064d93-257d-4c65-b173-3361141550d5](https://github.com/user-attachments/assets/c8f4128c-37cb-451e-85ce-35f13e5906a7)

## System Specs

OS: Arch Linux KDE Plasma X11/Wayland

## Workarounds

* Hooking up BSM with steam as a _non-steam game_, as discussed in #583
* Renaming the folders manually (or by a script) to remove any unicode characters from the song name.

## Additional context

The main issue here is within proton/wine itself when windows path is converted to linux path, but I'm not sure why running it as a _non-steam game_ solves this issue. Might try to look into environment variables that might lead to somewhere.

kevin-wijnen:
Hmmm, can not replicate it myself:

![image](https://github.com/user-attachments/assets/2866faec-6a92-488b-864c-2c734c66b99f)
![image](https://github.com/user-attachments/assets/dc6ea250-43c1-4ac4-aeb8-965725dfc73e)

On what filesystem did you try installing the songs onto, or is BSManager located in @silentrald? It could be related to the file system of the storage medium. I'm running `ext4` as the filesystem.

silentrald:
@kevin-wijnen It should show in the filesystem just fine. I just can't see the song in game when you run the AppImage within Dolphin, can you search it in game?

kevin-wijnen:
> @kevin-wijnen It should show in the filesystem just fine. I just can't see the song in game when you run the AppImage within Dolphin, can you search it in game? Also I forgot that this happens in 1.37.1 but not in 1.29.1.
> 
> Also I run `btrfs`.

My bad! Yeah, running 1.37.1 causes the entire custom song list to simply not load (looping infinitely). I could not find where the logs are, produced by the Debug Mode, however.

silentrald:
Ohhh that issue relates to #578, there were hidden dependencies that weren't documented so your BSIPA isn't run correctly but already gotten a #586 PR to remove those dependencies, since that's better for the Steam Deck users.

[...]

PR #586, opened 2024-09-22
[feat-578] remove dotnet and screen dependency on linux

silentrald:
## Scope

closes https://github.com/Zagrios/bs-manager/issues/578

## Implementation

Removed `dotnet` and `screen` dependencies. Then used `proton` and `wine` binary files which comes package in Steam. They should be linked with the **protonFolder** which is stored in the static config.

On startup as well, there is a check if the protonFolder is correctly configured, else the user will be blocked by a modal to configure the protonFolder. 

## Screenshots

image

image

## How to Test

* Setup
  * Prompting the linux setup modal
    * Remove the **proton-folder** key within `config.json`.

* Feature Tests
  * Check if setting the protonFolder to an invalid value is handled properly, both on the modal and settings page
  * Check if setting the protonFolder to a valid value is correct, both on the modal and settings page

* Functionality Tests
  * Check if you can install mods (especially BSIPA) within anyBSVersion
  * Check if you can launch any BSVersion

## New Resource Text

<details>
<summary>List of added resource text</summary>

**misc.unknown:** An unknown error occurred ¯\\_(ツ)_/¯

**pages.settings.proton-folder.title:** Proton folder
**pages.settings.proton-folder.description:** Change the folder to the Proton path. (eg. Proton - Experimental)
**pages.settings.proton-folder.choose-folder:** Choose folder
**pages.settings.proton-folder.errors.title:** Setting the Proton folder failed
**pages.settings.proton-folder.errors.invalid-folder:** Invalid Proton folder path

**modals.linux-setup.title:** Linux Setup
**modals.linux-setup.proton-folder:** Proton folder
**modals.linux-setup.proton-folder-description:** Choose the folder to the Proton path. (eg. Proton - Experimental)
**modals.linux-setup.choose-folder:** Choose folder

**notifications.bs-launch.errors.PROTON_NOT_SET:** Proton folder not set
**notifications.bs-launch.errors.PROTON_NOT_FOUND:** Proton binary not found
**notifications.bs-launch.msg.PROTON_NOT_SET:** Set the Proton path in settings
**notifications.bs-launch.msg.PROTON_NOT_FOUND:** Set the Proton path in settings

</details>


Zagrios:
@silentrald Do you know if there is a guide or something created by Steam that explains what Proton is to users? (What it is, where to find it, etc)

silentrald:
@Zagrios I think it's in the Proton's GitHub page.

What - https://github.com/ValveSoftware/Proton
Where - https://github.com/ValveSoftware/Proton/wiki/Proton-FAQ#where-is-proton-installed

Although I could defaultly open the `$HOME/.local/share/Steam/steamapps/common` on the folder modal, for a faster navigation of the proton binary.

Is there anything that should be added if there are any non-technical linux users that may come upon this modal screen?

---

Might add some additional changes, to use the user's `wine` executable if its installed in the users machine, else fallback back to the wine bundled with proton.

Also sometimes the `Beat Saber.exe` isn't found so we have to explicitly tell where it is. As discussed in https://github.com/Zagrios/bs-manager/issues/585#issuecomment-2367253761.

Zagrios:
@silentrald I was testing on my Ubuntu machine, and I couldn't get the BSIPA installation to work. 
So, I tried running that command in the terminal to see what would happen, and it turns out that Wine takes over 5 minutes to start `IPA.exe`. Is that normal behavior?
```
'/home/mathieu/.local/share/Steam/steamapps/common/Proton - Experimental/files/bin/wine64' '/home/mathieu/BSManager/BSInstances/1.37.5/IPA.exe' '/home/mathieu/BSManager/BSInstances/1.37.5/Beat Saber.exe' -n
```

And also for me, it only works with `wine64`; otherwise, using `wine` gives me a `kernel32.dll not found` error. :thinking:

silentrald:
@Zagrios normally the first call to wine, assuming that your `/.wine` environment isn't initialized yet, should download the necessary things for it work which makes it slow. Then for subsequent calls on wine should be fast. What I think that happened as well is that you ran the mod install which the spawn call timed-out, meaning that the wine process is not terminated in the system. Then you ran another instance of wine, which is getting blocked or waiting for the original wine process to be done. So I think you just needed to kill the wine process.

For the wine (32 bit) fallback, on Arch it works normally but in my Ubuntu VM, seems I'm encountering the kernel32.dll as well. Seems the problem lies in some dll dependency files that are not available in Ubuntu. However, wine64 seems to work there so I might change it to that and since electron-builder is only limited to `--x64` arch as well unless there is support for x86 for linux with that argument.

Also ran some problem with the installed wine with Ubuntu since this needs to have a `wine-mono` installed, which can be solve by this https://askubuntu.com/a/992215. Can add this to the wiki troubleshooting part so that it would be easier to install. Although, side note on Arch, wine ask you to install missing dependencies which is when you run the wine command, that's why its very weird that Ubuntu doesn't do this.

So which `wine` implementation do you think is the best here:
- **(My preffered option)** Use only proton `wine/wine64` binary depending on the architecture. Or we can just support wine64.
- Check if `wine` is installed and do the troubleshooting if necessary, if no wine is installed fallback to proton `wine/wine64`
- Use proton `wine/wine64` then fallback to installed `wine`.

Zagrios:
I think I will support only the proton/wine64

silentrald:
I actually didn't notice this but Ubuntu installs wine 9.0 in their stable release while in my build it uses 9.20. Followed this `winehq` tutorial (https://gitlab.winehq.org/wine/wine/-/wikis/Debian-Ubuntu) on how to upgrade it to 9.20 and BSIPA works, just have to adjust the spawn timeout or remove it, since wine-mono install takes a long time to download. This might be better since if wine updates to their next stable release, installing wine OOTB would have no more issues.

Might follow this flow instead:
- Run `wine --version` and check if its version
  - **>= 9.20** use installed wine, if does not exist fallback to proton wine64
  - **< 9.20** use proton wine64

This also details this command to support for 32 bit machines
```bash
sudo dpkg --add-architecture i386 
```

---

Wdyt, or just always use proton wine64 would be the better?

Zagrios:
Hmm, I think always using Proton/Wine64 would be better. This approach would reduce a bit the complexity and therefore minimize potential points of failure

"best code is no code" 😅

We know Proton/Wine64 works well, so that should be enough.

Zagrios:
Okay I've re-tested everything, and everything was working perfectly :raised_hands:
Thanks @silentrald !
Merging :rocket:

Zagrios (approved):

PR #589, opened 2024-09-23
[bugfix] show bsm icon properly

silentrald:
## Scope

Correctly point to icon images for both windows and linux

## Emoji Guide

> 🟢 Nice refactor!


Zagrios (approved):
Great, thanks ! 🙌

PR #596, opened 2024-09-30
[chore] improved user experience for bug reporting

silentrald:
_**Still subject to change, depending on the labels and inputs needed**_

image

NOTE: This might be better for the labeler to assign platform/os specific labels. Just need to search Windows/Linux within the issue body.


Zagrios (approved):
That's awesome 🤩! I didn't know that we could use yaml here for that kind of stuff 🤯

PR #599, opened 2024-10-02
[chore] properly add correct label for linux

silentrald:
Used `regex matching` instead of github actions simple `contains` check.


Zagrios (approved):
Thanks ! 🙌

Issue #608, opened 2024-10-12
[BUG] : Can not connect to Steam


FalsePsyche:
### Issue encountered

on 1.4.8 and 1.5 alpha 4, BSManager doesn't connect to Steam. Steam is installed and running on a windows 11 PC. Beat Saber 1.38 is installed. I've never used BSManager before, 1.4.8 is the first time I've installed it. When I open the app and select Beat Saber 1.38 (recommended) I get the steam login pop up. The QR code never renders, shown here 
![image](https://github.com/user-attachments/assets/020a18b5-7571-4bb5-b6ce-1c70f0151b5c)
When I put in my credentials I get the error can not connect to steam. 

BSManager works once I connect to a VPN. To be clear, without the VPN the windows Steam client works just fine.
Logs are attached [2024-10-12-v1.4.8.log](https://github.com/user-attachments/files/17350657/2024-10-12-v1.4.8.log)

### Operating System

None

Zagrios:
Hi, a patch is coming to fix that 👌

PR #609, opened 2024-10-13
[bugfix] update DepotDownloader to fix connection to steam errors

Zagrios:


silentrald:
@Zagrios working perfectly with linux 👌

Can also close issue #608 once this gets merged.

PR #613, opened 2024-10-15
[bugfix] delete map folders with invalid/corrupted info.dat files

silentrald:
image

## Test
1. Edit a folder's `info.dat` file into any text that isn't json format.
2. Delete the `song-raw-info-cache.json` file in the bs-manager cache folder.
3. Open the version that contains the map.
4. Verify if the log is shown and the folder is deleted correctly.


silentrald:
Closing PR. Handled nicely with 7468554ea03628db7a96834918c37d3a7ee81683.

Although might be a good idea that when any map folders are corrupted/invalid, we can ask the user to choose to delete or retain the map folder in the system. We can also add a checkbox that says `don't show this again` so that the user isn't spammed with deletion every time they open the app.

PR #616, opened 2024-10-16
[bugfix-614] handle bplist source string within readPlaylistFromSource

### LINKED ISSUE #614: [BUG] : Playlist created in BSM does not work in linux
### Issue encountered

image

When trying to sync created playlist with BSM, the above error occurs. Could not verify if this is the same on Windows.

### Expected Behavior

Playlist should sync correctly.

### Steps to Reproduce

1. Create a playlist with BSM.
2. Sync the created playlist.
3. Error should prompt on the renderer and the console.

### Operating System

Linux

### Version

1.5.0-alpha.4

### Additional Context

Not sure on how the whole playlist gets its file but dug a little bit of info that the fetch command tries to query it twice. One with the `file:///` and another without it. Not sure if that is expected, might do a little bit of research on this.

<pre>
/home/some/path/to/sample.bplist
file:///home/some/path/to/sample.bplist
</pre>

silentrald:
Closes #614

* refactored to just use bplistSource each time it gets passed into another function
* removed unused songCache variable

NOTE: not yet tested on Windows

### DIFF
--- a/src/main/ipcs/bs-playlist-ipcs.ts
+++ b/src/main/ipcs/bs-playlist-ipcs.ts
@@ -3,8 +3,6 @@ import { IpcService } from "../services/ipc.service";
 import { from, lastValueFrom, mergeMap, of } from "rxjs";
 import { LocalMapsManagerService } from "../services/additional-content/maps/local-maps-manager.service";
 import { Progression } from "main/helpers/fs.helpers";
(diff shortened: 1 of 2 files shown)


Zagrios (approved):
Thanks! Tested on Windows, it works 👌
I had forgotten that `readPlaylistFromSource` was already checking if it's from a local file or not 😅

PR #641, opened 2024-11-04
[Transation] Add Korean Languages

izunya:
https://github.com/user-attachments/assets/db6ed113-0559-4a01-9eaa-a38858be195e

I'm not skilled enough to create a flag icon. I would really appreciate it if you could make one for me.
image

PR #644, opened 2024-11-05
[Bugfix&Add-#641] Added kr translations | Text wrap bug and incorrect translation of "Curated"

izunya:
#641

# Before
image
image
image
image

# After
image
image
image
image


Zagrios (approved):
Nice thanks ! 🙌

PR #649, opened 2024-11-07
[bugfix] handle extracting non-conformant zip files

silentrald:
Issue described in [discord](https://discord.com/channels/1049624409276694588/1303885183094751262)

- [x] Change extract logic in import maps.
- [x] Remove `jszip` and `node-stream-zip` packages.

**NOTES:**
- Used `yauzl` since that support non-conformant zip files and ram efficiency.
  - Caveats, may slow down zip extraction a little bit.
- Added test on standard and the breaking map zip file.
- Removed `jszip` in getting oculus manifest and I can't test it on my end.
- Ready for review <s>but will need to address some comments (BUG and NOTE)</s>.
- Do we need to add more tests on big zip files, or a sample exported map files to see the memory usage?


silentrald:
@Zagrios yeah its a little bit over 😅 but was trying to not expose the yauzl module to everything so that its testable, so whenever a change happens, its should be isolated within zip.helpers.

Although I encountered that in coding and I suspect its in the `Subject` that I was using in importMaps. Sometimes it would drop `next` events and when that happens, the **current != total** occurs. I tested on mine and works with 89 songs but ill try to exhaustively add more songs to check if I can replicate.

silentrald:
Tested on around ~1,400 maps and its working fine 🤔 so its very weird that its failing

silentrald:
*Notes:*
* Never though of generators 😯 but seems to be working, only had to set `autoClose: false` to make everything work as expected.
* Just a weird bug with `path.matchesGlob` doesnt support escaping special characters, just used regex for this instead.
* Maybe consider changing the name of YauzlZip to another name class like BsmZip since in theory, we own/control the function calls on that class.

Zagrios (approved):
🙌

Zagrios:
Thanks for the help @silentrald ! ❤️ 
Everything's working fine 👌 
Merging 🚀

PR #652, opened 2024-11-09
[bugfix-452] fix to "launch game when steam is not running" on linux

### LINKED ISSUE #452: [BUG] : Steam doesn't get started if it's not already running on Linux
## Bug Description
When launching Beat Saber on Linux, if Steam isn't already running, it doesn't get launched. The launch attempt is made, but it times out without Steam ever starting.

## Reproduction Steps
- Ensure Steam is closed.
- Launch Beat Saber.
- Observe Steam not launching.

## Expected Behavior
Steam should be launched.

## System Specs
OS: Arch Linux
Version: 9cd5c488310ab83fa1601970f38d43b2b905044f

LiamillionSS:
Closes #452 

Fixes issue of steam not launching when it is not already open. Also flatpak now launches Beat Saber as expected

### DIFF
--- a/src/main/services/steam.service.ts
+++ b/src/main/services/steam.service.ts
@@ -13,7 +13,9 @@ const { list } = (execOnOs({ win32: () => require("regedit-rs") }, true) ?? {}) 
 
 export class SteamService {
 
-    private static readonly PROCESS_NAME = "steam";
+    private static readonly PROCESS_NAME: string = process.platform === "linux"
+        ? "steam-runtime-launcher-service"
+        : "steam";
 
     private static instance: SteamService;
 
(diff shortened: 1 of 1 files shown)


silentrald:
@LiamillionSS I think there is also a bug in detecting whether or not steam is running. It might return false positives when lets say you open a text file within the steam folder and the resulting process name looks like this.

image

I suggest to change SteamService.PROCESS_NAME to "steam-runtime-launcher-service" instead of "steam". This should only be for "linux" and shouldn't affect windows

Zagrios (approved):
Clean, simple, perfect :ok_hand: Thanks!

Zagrios:
Tested on Windows and Linux, it works perfectly fine :ok_hand: 
Thanks @LiamillionSS ! :heart: 
Merging :rocket:

LiamillionSS:
HeroicGamesLauncher had the same issue, open-source software never fails to deliver

PR #662, opened 2024-11-18
Set default beat saver map sort order to Relevance

Top-Cat:
## Scope

Change default sort order of beatsaver maps to Relevance

Obviously you want to see Latest maps initially but fun fact: If you don't supply a search parameter Relevance returns the latest maps.

It's a really awful user experience to see the Latest maps when searching for something unless you've specifically asked for that:
(I know these aren't app screenshots but it's the same search results)
Screenshot 2024-11-18 at 16 28 49
Screenshot 2024-11-18 at 16 28 59


Zagrios (approved):
🙌

Zagrios:
Great! Thanks a lot @Top-Cat ❤️

PR #664, opened 2024-11-19
[feat/663] Option to launch BS without SteamVR

### LINKED ISSUE #663: [FEAT.] : Option to launch BS without SteamVR
### Problem

SteamVR conflicts with with other VR solutions like WiVRn or Monado (or ALVR to some extent), so an option to launch BS straight away without VR=Running checks would bypass a lot of issues

### Solution

IMO would be included in the advanced settings as this is a niche case that wont be used for many installs
image

### Alternative Solution

_No response_

### Operating System

None

### Additional Context

_No response_

LiamillionSS:
Closes #663, which is to add a button to the Advanced Settings that enables and disables the "openSteam" part of the launch functionality when toggled (+ minor typo)

Set to enable launching as default to match original configuration

Once wording is finalised I will add the rest of the translations


src/main/services/bs-launcher/steam-launcher.service.ts
```diff
@@ -77,8 +77,11 @@ export class SteamLauncherService extends AbstractLauncherService implements Sto
                 throw CustomError.fromError(new Error(`Path not exist : ${exePath}`), BSLaunchError.BS_NOT_FOUND);
             }
 
+            const staticConfigService = StaticConfigurationService.getInstance();
+            const launchVREnabled = await staticConfigService.get("disable-launch-vr");
```
silentrald, line 81: From a readability standpoint, this is very confusing. If "disable-launch-vr" is true then this will check if steam is running. Might want to use "enable-launch-vr" instead, or "enable-steamvr-launch" that other dev may get context that this is tied to SteamVR and Steam.

silentrald, line 81: Remove the `staticConfigService` variable and use the `this.staticConfig` member variable instead.

LiamillionSS: I did that originally but changed it to match the formatting of hardware acceleration 💀

Alright that's easy enough to redo

Zagrios:
@LiamillionSS How about putting that in the advanced launch panel instead? 🤔
I mean, it’s kind of made for that😅
image

Zagrios:
@LiamillionSS Could you replace "Skip SteamVR" with "Skip Steam"? 🤔
'Cause from what I see in the code, it doesn't prevent SteamVR from opening but rather skips starting Steam before launching Beat Saber 🤔

src/main/services/bs-launcher/abstract-launcher.service.ts
```diff
@@ -37,6 +37,9 @@ export abstract class AbstractLauncherService {
         if (launchOptions.additionalArgs) {
             launchArgs.push(...launchOptions.additionalArgs);
         }
+        if (launchOptions.launchVR) {
+            launchArgs.push(...launchOptions.launchVR);
```
Zagrios, line 41: @LiamillionSS is this still relevant 🤔 ?

Zagrios (changes requested):

Zagrios:
Thanks @LiamillionSS ❤️ 
Merging 🚀

PR #674, opened 2024-11-26
wiki: easier method to obtain oculus token

basti564:
Replaced the wiki page on how to obtain an oculus token with a guide that is straight forward, only needs a web browser and doesn't require tinkering with the Oculus PC app :3


basti564:
> Tokens starting with "OC" have stopped working a long time ago for downloading versions (since this PR was closed: #383) Previously, we had a system to automatically retrieve this cookie using an embedded browser

The token starting with "OC" doesn't mean anything for its capabilities
You should try again with the token from the `oc_ac_at cookie`. It's scoped correctly to be able download builds you are entitled to without being the developer :3
Old methods used to use the `oc_www_at` access token which doesn't have enough permission to do so

Zagrios:
@basti564 thanks for the info! I tested it, and the `oc_ac_at` cookie works as expected
I'll probably reimplement the feature to log in to Oculus directly from BSManager 👌 

I'll merge your PR soon, just need to adjust some things in BSM to support OC tokens 😉

Zagrios (approved):
Looks good to me 🙌

Zagrios:
Thanks a lot @basti564 ! ❤️ 
Merging 🚀

PR #685, opened 2024-12-10
[bugfix] remove auto update for linux builds

silentrald:
Since linux users update their application thru the package manager, disabled auto updates thru BSManager to not conflict or create issues with updates.

image

### DIFF
--- a/src/renderer/components/title-bar/title-bar.component.tsx
+++ b/src/renderer/components/title-bar/title-bar.component.tsx
@@ -12,6 +12,51 @@ import { useService } from "renderer/hooks/use-service.hook";
 import { lastValueFrom } from "rxjs";
 import { useWindowControls } from "renderer/hooks/use-window-controls.hook";
 
+function TitleBarTags({
+    ipcService
+}: Readonly<{
+    ipcService: IpcService
+}>) {
+    const [previewVersion, setPreviewVersion] = useState("");
+    const [outdated, setOutdated] = useState(false);
+
+    useEffect(() => {
+        const requests: Promise<any>[] = [
+            lastValueFrom(ipcService.sendV2("current-version"))
+        ];
+        if (window.electron.platform === "linux") {
+            requests.push(lastValueFrom(ipcService.sendV2("check-update")));
+        }
+
+        Promise.all(requests).then(([ currentVersion, outdated ]) => {
+            handlePrerelease(currentVersion);
+            setOutdated(outdated);
+        });
+    }, []);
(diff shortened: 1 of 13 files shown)


src/renderer/components/title-bar/title-bar.component.tsx
```diff
@@ -12,6 +12,51 @@ import { useService } from "renderer/hooks/use-service.hook";
...
+
+    return <>
+        {previewVersion &&
+            <span className="bg-main-color-1 text-white dark:text-black dark:bg-white rounded-full ml-1 text-[10px] italic px-1 uppercase h-3.5 font-bold">
+                {previewVersion}
+            </span>
+        }
+        {outdated &&
+            <span className="bg-main-color-1 text-white dark:text-black dark:bg-white rounded-full ml-1 text-[10px] italic px-1 uppercase h-3.5 font-bold">
+                Outdated
```
Zagrios, line 54: A missing translation?

src/renderer/components/title-bar/title-bar.component.tsx
```diff
@@ -76,7 +108,7 @@ export default function TitleBar({ template = "index.html" }: { template: AppWin
                 <div id="drag-region" className="grow basis-0 h-full">
                     <div id="window-title" className="pl-1">
                         <span className="text-gray-800 dark:text-gray-100 font-bold text-xs italic">BSManager</span>
-                        {previewVersion && <span className="bg-main-color-1 text-white dark:text-black dark:bg-white rounded-full ml-1 text-[10px] italic px-1 uppercase h-3.5 font-bold">{previewVersion}</span>}
+                        <TitleBarTags ipcService={ipcService} />
```
Zagrios, line 111: There is a reason to pass the service as a props, instead of using the `useService` hook directly in the component?

Zagrios (changes requested):

src/renderer/components/title-bar/title-bar.component.tsx
silentrald: Ohh yeah, forgot it was a singleton. Might be better to do that then.

src/renderer/components/title-bar/title-bar.component.tsx
Zagrios: `Outdated` should do the work
For the color, try to use the `bg-warning-xxx` for the background color (adjust the text color if needed)

Zagrios (approved):

PR #686, opened 2024-12-10
[bugfix] change linux shell from /bin/bash to bash

silentrald:
Just used default `bash` rather than absolute `/bin/bash` to get the correct shell from all distros.


Zagrios (approved):
Great 🙌

PR #691, opened 2024-12-11
[feat] added blacklist for shared folders

silentrald:
* change BasicModal data to handle cascading modals better
* "junction" is only supported in windows as documented in https://nodejs.org/api/fs.html#fspromisessymlinktarget-path-type. Disabled symlink settings for non-windows OS'es.

**Warning**
image

**Error**
image


Zagrios (approved):
Everything seems good to me 🙌

Zagrios:
Thanks @silentrald ❤️ 
Merging 🚀

Issue #702, opened 2024-12-17
[FEAT.] : Follow system proxy setting


GoldJohnKing:
### Problem

Could you make BSManager use system proxy on Windows?

In my region (China Mainland), I cannot connect to BeatSaver without a proxy.

BSManager currently does not follow system proxy settings, which means all online features (such as one-click map install) is not functional. 🥹

### Solution

Follow system settings by default, or provide an option to let user choose if using system proxy.

### Alternative Solution

Currently, I can force a global proxy to make BSManager work.

However, global proxy must capture all network traffic, which reduces overall network performance.

This causes severe stutters in streaming applications such as SteamLink or VirtualDesktop.

That means, I cannot use BSManager and SteamLink at the same time.

### Operating System

Windows

GoldJohnKing:
Hi, I dig a bit.

BSManager uses `got` as its request handler, which is supported by [`global-agent`](https://www.npmjs.com/package/global-agent).

[`os-proxy-config`](https://www.npmjs.com/package/os-proxy-config) could be useful for getting system proxy settings.

However, I am really not familiar with NodeJS, or I could implement that myself... 🥹

PR #704, opened 2024-12-18
feat: respect system proxy settings on Windows platform

GoldJohnKing:
## Scope

[closes TICKET-702](https://github.com/Zagrios/bs-manager/issues/702)

## Implementation

Makes BSManager respect system proxy settings on Windows.

1. Detect if BSManager is running on Windows. If it does, execute following steps.
2. Read Windows system proxy settings from registry by using `Winreg` package.
3. Force a global proxy by using `global-agent` package.

I am really not familiar to NodeJS, so implementation code should be ugly. Feel free to edit the code or give any suggestions.

I do not have Linux and macOS and is unlikely to use them. This implementation does not cover Linux or macOS.

## How to Test

Without system proxy enabled on Windows:

- Run BSManager, all functions works fine like before.
- In the log of BSManager, you should see `Proxy Enabled: false`.

With system proxy enabled on Windows:

- Set system proxy settings and enable it on Windows.
- Run BSManager.
- In the log of BSManager, you should see:
  ```log
  [2024-12-18 17:08:04.386] [info]  Proxy Server:  http://127.0.0.1:7890
  [2024-12-18 17:08:04.403] [info]  Proxy Override:  localhost;127.*;192.168.*;10.*;172.16.*;172.17.*;172.18.*;172.19.*;172.20.*;172.21.*;172.22.*;172.23.*;172.24.*;172.25.*;172.26.*;172.27.*;172.28.*;172.29.*;172.30.*;172.31.*;<local>
  [2024-12-18 17:08:04.403] [info]  Using system proxy: http://127.0.0.1:7890
  ```
  That means BSManager is using system proxy.
- If your have access to the proxy and can see which traffic goes through it, you can see BSManager's traffic goes through it.

## Emoji Guide

> 🟢 Nice refactor!


silentrald:
@GoldJohnKing
Can you try if you can make this toggle-able in the settings page. So that users can just turn this on if they need to when they have proxy issues.

Is it also possible to set the proxy string within BSManager itself as well 🤔?

---

Also not sure if `Winreg` is just viewing the Regedit values but can you check `regedit-rs` since its similar and already within the project's dependencies, much better to reuse existing deps rather than adding one.

Zagrios (changes requested):
Hi, thanks a lot for your PR!
But can you use `regedit-rs`, which we already use for registry manipulation instead of `winreg`?
https://www.npmjs.com/package/regedit-rs

GoldJohnKing:
Hi, I made the changes from `winreg` to `regedit-rust`.

As for the toggole-able setting page option suggested by @silentrald, I am not familiar to NodeJS and Electron. If it's a necessary function, would you please point me where to start from?

silentrald:
@GoldJohnKing 

Try looking into `advancedItems` in `settings-page.component.tsx`. Then you can use staticConfig set and get from the client and server side code. We can just do the toggle input for now I guess, but if you can add the input text then it a nice to have feature.

Here's the reference in BSM itself.
image

GoldJohnKing:
@silentrald @Zagrios 

Hi, toggle-able option is added, please check.

src/main/helpers/proxy.helpers.ts
```diff
@@ -0,0 +1,75 @@
+import log from 'electron-log';
```
silentrald, line 1: Can this logic be in `request.service.ts` instead since `proxy.helper.ts` is just only used by that specific singleton class, no point of exposing proxy stuff outside request/network code.

src/renderer/pages/settings-page.component.tsx
```diff
@@ -609,6 +611,45 @@ function AdvancedSettings() {
...
+        }
+
+        setUseSystemProxy(() => newUseSystemProxy);
+
+        if(!progressBar.require()){
+            return;
+        }
+
+        if (!newUseSystemProxy) {
+            await lastValueFrom(ipc.sendV2("restart-app"));
```
silentrald, line 648: Does the app need to be restarted in order for the change to kick in? If I'm reading how the proxy enabling works, its just setting the `process.env.GLOBAL_AGENT_HTTP/S_PROXY` right? So in theory you can just delete the keys so that the proxy is disabled. Can you double check if thats the case.

assets/jsons/translations/en.json
```diff
@@ -279,6 +279,18 @@
...
+                    "title": "Use system proxy",
+                    "description": "BSMansger will make network requests through system proxy.",
+                    "modal": {
+                        "title": "Restart Needed",
+                        "body": "Turning off system proxy will quit and re-launch BSManager. Are you sure you want to do this?",
+                        "confirm-btn": "Yes I'm sure"
+                    },
+                    "error-notification": {
+                        "message": "An error occured, unable to use system proxy."
+                    }
```
silentrald, line 293: Some update to the text
<pre>
title: Use System Proxy
description: BSManager will make network requests through your system proxy.
message: An error occurred, unable to change system proxy settings.
</pre>

Also check if you could do translation for other language, we use gpt-4 for the translation btw.

---

Also as a favor, `occurred` is misspelled on the other settings as well, can you fix them in your next commit, thanks thanks.

silentrald (changes requested):

src/renderer/pages/settings-page.component.tsx
GoldJohnKing: > Does the app need to be restarted in order for the change to kick in? If I'm reading how the proxy enabling works, its just setting the `process.env.GLOBAL_AGENT_HTTP/S_PROXY` right? So in theory you can just delete the keys so that the proxy is disabled. Can you double check if thats the case.

I tried that, empty the keys do not disable the proxy agent. You may give another try to see if it works.

src/main/helpers/proxy.helpers.ts
GoldJohnKing: Personally I would prefer making it as standalone as possible.

As of its concept, it is much closer to a system specific helper module like `hardware acceleration toggle`, instead of a sub-module of `request`. Besides, `request` does not require `proxy` to work.

Since it does not only affect network code in `request`, I would prefer moving the entry to `main.ts`, just below `hardware acceleration`. I think it's much better than invading/injecting it into `request`.

GoldJohnKing: Hi, I can confirm moving the entry to `main.ts` solves the following two connection failure due to network issue.

It looks like a better solution than making it initialize in `request`.

```log
[2024-12-24 21:28:45.127] [warn]  Download failed over IPv4 for URL: https://raw.githubusercontent.com/Zagrios/beat-saber-scraped-maps/master/song_details_cache_v1.gz RequestError: read ECONNRESET
    at ClientRequest.<anonymous> (D:\GitRepos\bs-manager\release\build\win-unpacked\resources\app.asar\dist\main\main.js:14:1629331)
    at Object.onceWrapper (node:events:634:26)
    at ClientRequest.emit (node:events:531:35)
    at ClientRequest.emit (node:domain:488:12)
    at emitErrorEvent (node:_http_client:108:11)
    at TLSSocket.socketErrorListener (node:_http_client:511:5)
    at TLSSocket.emit (node:events:519:28)
    at TLSSocket.emit (node:domain:488:12)
    at emitErrorNT (node:internal/streams/destroy:169:8)
    at emitErrorCloseNT (node:internal/streams/destroy:128:3)
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:218:20)
```

```log
[2024-12-24 22:07:09.251] [warn]  IPv4 request failed, trying next one... URL: https://raw.githubusercontent.com/Zagrios/bs-manager/master/assets/jsons/bs-versions.json RequestError: connect ETIMEDOUT 185.199.110.133:443
    at ClientRequest.<anonymous> (D:\GitRepos\bs-manager\release\build\win-unpacked\resources\app.asar\dist\main\main.js:14:1586333)
    at Object.onceWrapper (node:events:634:26)
    at ClientRequest.emit (node:events:531:35)
    at emitErrorEvent (node:_http_client:108:11)
    at TLSSocket.socketErrorListener (node:_http_client:511:5)
    at TLSSocket.emit (node:events:519:28)
    at emitErrorNT (node:internal/streams/destroy:169:8)
    at emitErrorCloseNT (node:internal/streams/destroy:128:3)
    at processTicksAndRejections (node:internal/process/task_queues:82:21)
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1607:16)
[2024-12-24 22:07:09.260] [warn]  IPv6 request failed, trying next one... URL: https://raw.githubusercontent.com/Zagrios/bs-manager/master/assets/jsons/bs-versions.json RequestError: getaddrinfo ENOENT raw.githubusercontent.com
    at ClientRequest.<anonymous> (D:\GitRepos\bs-manager\release\build\win-unpacked\resources\app.asar\dist\main\main.js:14:1586333)
    at Object.onceWrapper (node:events:634:26)
    at ClientRequest.emit (node:events:531:35)
    at emitErrorEvent (node:_http_client:108:11)
    at TLSSocket.socketErrorListener (node:_http_client:511:5)
    at TLSSocket.emit (node:events:519:28)
    at emitErrorNT (node:internal/streams/destroy:169:8)
    at emitErrorCloseNT (node:internal/streams/destroy:128:3)
    at processTicksAndRejections (node:internal/process/task_queues:82:21)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:109:26)
[2024-12-24 22:07:09.260] [error] IPv4 and IPv6 requests failed for URL: https://raw.githubusercontent.com/Zagrios/bs-manager/master/assets/jsons/bs-versions.json
```

src/renderer/pages/settings-page.component.tsx
silentrald: Got to enable/disable toggling working without restarting the app. Read the docs that you can edit the `global.GLOBAL_AGENT` object in runtime once `bootstrap` setup is called. So I did the below code:

```typescript
// Changed `configureWindowsProxy` to `enableWindowsProxy`
async function enableWindowsProxy(enable: boolean): Promise<void> {
    if (!(await isProxyEnabled().catch(err => log.error(err)))) {
        log.info(`configureWindowsProxy: System proxy not detected`);
        return;
    }

    let { GLOBAL_AGENT: globalProxyAgent } = global as any;
    if (!globalProxyAgent) { // If this is undefined, call bootstrap to set it up
        if (!bootstrap()) {
            log.error(`configureWindowsProxy: Could not setup proxy stuff`);
            return;
        }
        globalProxyAgent = (global as any).GLOBAL_AGENT;
    }

    if (enable) {
        const httpProxyUrl = `http://${await getProxyServer().catch(err => log.error(err))}`
        globalProxyAgent.HTTP_PROXY = httpProxyUrl;
        globalProxyAgent.HTTPS_PROXY = httpProxyUrl;
        globalProxyAgent.NO_PROXY = `${await getProxyOverride().catch(err => log.error(err))}`;
    } else {
        delete globalProxyAgent.HTTP_PROXY;
        delete globalProxyAgent.HTTPS_PROXY;
        delete globalProxyAgent.NO_PROXY;
    }

    log.info(`configureWindowsProxy: Using system proxy: ${globalProxyAgent}`);
}
```

That should be the core of the change, you can handle the rest of the change at this point.

assets/jsons/translations/en.json
GoldJohnKing: Done.

src/renderer/pages/settings-page.component.tsx
GoldJohnKing: Thanks! Done.

assets/jsons/translations/en.json
GoldJohnKing: Done.

GoldJohnKing:
@silentrald 

Hi, all requested changes have been done, please check.

Thanks for your suggests!

Zagrios (approved):
That's seems good to me 🙌 
Thanks @GoldJohnKing ❤️

Zagrios:
Thanks @silentrald for the review ❤️ 
Github tells me their is still one requested change pending but I didn't see it anywhere, is it a ghost request changes ? 😅

silentrald:
@Zagrios let me double check if i can unblock it on my end

silentrald (approved):

PR #710, opened 2024-12-22
[chore] added instruction for adding logs in issue template

silentrald:
image
NOTE: Not sure if adding a separate field for log files is better than just having it being dropped in the addition content textarea field

### Disabled submitting blank issues

**Rationale:** so we can force users to use the templates which has also have instruction on adding log files to reduce issue turn over time if possible.


Zagrios:
Thanks @silentrald 🙌 
Merging 🚀

Zagrios (approved):

PR #712, opened 2024-12-25
[wiki] added deb install thru ppa repo in github

silentrald:
**Future Notes**
Maybe creating a github hook would be good as well to bump the deb version for each release after 1.5.0 rolls out.

### DIFF
--- a/docs/wiki/Linux.md
+++ b/docs/wiki/Linux.md
@@ -6,10 +6,33 @@ Go to [Releases](https://github.com/Zagrios/bs-manager/releases) page and go to 
 
 ### Ubuntu, Debian (deb)
 
+#### PPA Repository
+
+Refer to [bs-manager-deb](https://github.com/silentrald/bs-manager-deb).
+
+Add the BSManager PPA repository into your system with the following commands.
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/silentrald/bs-manager-deb/refs/heads/main/KEY.gpg | sudo gpg --dearmor -o /usr/share/keyrings/bs-manager.gpg
+echo "deb [signed-by=/usr/share/keyrings/bs-manager.gpg] https://raw.githubusercontent.com/silentrald/bs-manager-deb/refs/heads/main ./" | sudo tee /etc/apt/sources.list.d/bs-manager.list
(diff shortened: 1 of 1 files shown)


Zagrios:
@silentrald I see you've put silentrald/bs-manager-deb. No problem with that but does this mean you’ll need to update it manually for each new release of BSM?
What happens if you suddenly stop working on BSM (we never know)? 
Will it still be possible for us to create our own repository and switch to it so that the command `sudo apt install bs-manager` continues to work (if that happens)?

silentrald:
@Zagrios 

> silentrald I see you've put silentrald/bs-manager-deb. No problem with that but does this mean you’ll need to update it manually for each new release of BSM?

Yep. Although I was thinking if creating something like a hook with the bs-manager repo, similar to bots creating bump PRs and just commenting would approve it, to a degree automating it in someway. Can wait for 1.5.0 for it since this we're still in alpha anw.

> What happens if you suddenly stop working on BSM (we never know)?
> Will it still be possible for us to create our own repository and switch to it so that the command `sudo apt install bs-manager` continues to work (if that happens)?

In theory you could fork the repo and just update the keyring gpg and list. Since its the same deb package, in theory it would still work.

I could give you a collab role with bs-manager-deb so you can still bump the version yourself, or I could transfer the repo to you (or maybe a bsm organization 🤔). Which option would you prefer in any case?

Zagrios:
I would like to have the collab role for now then @silentrald 
I’ll look into creating an organization later. I’m wondering if moving the bs-manager repo to the organization might break the auto-update 🤔

silentrald:
> I would like to have the collab role for now then silentrald I’ll look into creating an organization later. I’m wondering if moving the bs-manager repo to the organization might break the auto-update 🤔

Sure sure, will send you the role in a bit. For the transfer, I might do research on it if auto updates break, but in theory you need to point the auto updater to the correct repo/url, not sure if it supports some kind of multiple url to pull new versions from.

Zagrios (approved):
Looks good to me 🙌

Zagrios:
Thanks a lot @silentrald ! ❤️ 
Merging 🚀

PR #720, opened 2024-12-29
[bugfix] fix oculus launch by using sideloading

Zagrios:
image


AltyFox:
@Zagrios please specify that this is a one-time permission, and won't be necessary in the future

Zagrios:
> @Zagrios please specify that this is a one-time permission, to permanently unlock the feature.

yep done 😉

AltyFox:
> > @Zagrios please specify that this is a one-time permission, to permanently unlock the feature.
> 
> yep done 😉

I like the wording, thank you!

AltyFox:
@Zagrios to maintain consistency with the rest of our utilities, there should be a way to teach the user how to manually do this in case they don't want to provide administrator access for whatever reason

Zagrios:
> @Zagrios to maintain consistency with the rest of our utilities, there should be a way to teach the user how to manually do this in case they don't want to provide administrator access for whatever reason

Yeah okay, I will add a button "I want to do it manually" or "How to do it manually" which will open a page on the BSM wiki 👌

PR #724, opened 2025-01-01
Update flatpak dependencies to non-EOL versions

eaglesemanation:
## Scope

electron-builder [defaults to 20.08](https://www.electron.build/flatpak.html#baseversion) version of flatpak runtime and base packages. Those packages are EOL, here is example message from flatpak cli:
```
Info: runtime org.freedesktop.Platform branch 20.08 is end-of-life, with reason:
   org.freedesktop.Platform 20.08 is no longer receiving fixes and security updates. Please update to a supported runtime version.
```
I couldn't really find a source of truth for those messages, I suspect it just part of flatpak repository that can report if package is EOL

## Implementation

Manually specify versions for both runtime and base package. This will require to manually bump them later on, but it's just 2 lines of code change approximately every 2 years, so I hope that's not an issue =)

## Screenshots

| before | after |
| ------ | ----- |
|        |       |

## How to Test

Try building flatpak package on a linux distro with `npm run publish:flatpak`

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/electron-builder.config.js
+++ b/electron-builder.config.js
@@ -51,6 +51,10 @@ const config = {
         fpm: ["--after-install=build/after-install.sh"],
     },
     flatpak: {
+        // Version of org.electronjs.Electron2.BaseApp
+        baseVersion: "24.08",
+        // Version of org.freedesktop.Platform
+        runtimeVersion: "24.08",
         finishArgs: [
             // Wayland/X11 Rendering
             "--socket=wayland",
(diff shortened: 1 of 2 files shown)


silentrald:
@eaglesemanation Seems to be good, could build the flatpak on my system.

@Zagrios could merge this PR. No workflow changes as well.

Zagrios (approved):
Seems good to me 🙌

Zagrios:
Thanks a lot @eaglesemanation ! ❤️ 
Thanks @silentrald for the review 🙌 
Merging 🚀

PR #726, opened 2025-01-02
[hotfix] fix issue where installed mods are not reflected in mods tab

silentrald:
Fixed an issue in the frontend where installedMods can contain a null element. Just removed the elements from the array.


Zagrios (approved):
Looks good to me 🙌

Zagrios:
Thanks @silentrald !
Merging 🚀

Issue #733, opened 2025-01-08
[BUG] : BSIPA install failed on Linux


zainin:
### Issue encountered

BSIPA fails to install on Linux with Beat Saber installation managed by Steam.

### Steps to Reproduce

Install BSManager and try to install BSIPA.

### Operating System

Linux

### Version

v1.5.0-alpha.7

### Additional Context

Looks like the WINEPREFIX isn't correctly set during the install so it tries to use the default prefix:
```
[2025-01-08 18:34:25.897] [info]  executeBSIPA 1.40.0 [ '-n' ]
[2025-01-08 18:34:25.898] [info]  START IPA PROCESS "/media/storage-ext/Games/SteamLibrary/steamapps/common/Beat Saber/IPA.exe" "/media/storage-ext/Games/SteamLibrary/steamapps/common/Beat Saber/Beat Saber.exe" -n
[2025-01-08 18:34:25.899] [info]  Linux spawn command
> "/media/storage-ext/Games/SteamLibrary/steamapps/common/Proton 9.0 (Beta)/files/bin/wine64" "/media/storage-ext/Games/SteamLibrary/steamapps/common/Beat Saber/IPA.exe" "/media/storage-ext/Games/SteamLibrary/steamapps/common/Beat Saber/Beat Saber.exe" -n
[2025-01-08 18:34:25.909] [error] IPA process stderr wine: using kernel write watches, use_kernel_writewatch 1.

[2025-01-08 18:34:25.910] [error] IPA process stderr wineserver: using server-side synchronization.

[2025-01-08 18:34:25.968] [error] IPA process stderr wine: '/home/zainin/.wine' is a 32-bit installation, it cannot support 64-bit applications.
```

As a workaround running BSIPA install manually with the correct prefix allows the installation to succeed.

silentrald:
@zainin Is your WINEPREFIX a custom value? Also when you run your inatalled wine version, did you also have to set this in the command?

zainin:
No, I use the prefix that's automatically set up by Proton, default Steam install.  
Normally I launch BS through Steam, and the manual install I did with `protontricks-launch`, so all the env vars are set up automatically.

But in general, the `WINEPREFIX` should be set to something like `$STEAM_DIR/steamapps/compatdata/620980/pfx`.

silentrald:
@zainin can you check if this build works on your machine and just update us if it works.
https://github.com/silentrald/bs-manager/actions/runs/12693493704

cc @Zagrios

[...]

PR #736, opened 2025-01-12
[bugfix-733] custom compatdata folder for specific BSManager usage

silentrald:
Link #733 

The compatdata folder can exist outside of the steam `compatdata` folder. 

In theory of this as well, BS versions can have their on specific configs that doesn't affect every BS version, such as the UserData folder. However the space consumption can increase since the compatdata folder contains the windows OS files/directories.

Notes: you can also launch the game without downloading the steam version of the game, only for linux. Not sure if this can be done on Windows.


Zagrios (approved):
Looks good to me 🙌

PR #749, opened 2025-01-20
[feat] inject env vars to launch options with steam format %command%

silentrald:
Issue described in [discord](https://discord.com/channels/1049624409276694588/1329541144836571197)

Notes:
- implemented a generic env parsers that is very similar to how steam handle theirs.
- also added generic error message that can be display generic error messages.


silentrald:
@Zagrios can you check if "Could not properly parse the env var string." would fine for env parsing errors, before I go translating. Also check my test cases for `parseEnvString` should also suffice, or we can add more to handle some error cases as well.

Zagrios (approved):
Looks good to me 👌

Zagrios:
Thanks @silentrald ! ❤️ 
Merging 🎉

PR #772, opened 2025-01-31
[feat] added logs when deleting files/folders

silentrald:
Issue from [discord](https://discord.com/channels/1049624409276694588/1334721659696971896)

When deleting a file or folder, log them so that it could be tracked in the logs file when checking for user errors.

Instead of using the direct function in fs, use there equivalent in fs.helpers module implementation.
unlink => deleteFile
unlinkSync => deleteFileSync
rm => deleteFolder
rmSync => deleteFolderSync

NOTE:
The wrapper function contains error codes without any resource, but they don't report anything on the frontend yet so didn't add it for now.


Zagrios:
Thanks @silentrald ❤️ 
Merging 🎉

PR #773, opened 2025-01-31
[feat] added auto update option on advance settings for windows

silentrald:
Feature described in [discord](https://discord.com/channels/1049624409276694588/1334667795358810204)


src/main/main.ts
```diff
@@ -132,10 +130,12 @@ if (!gotTheLock) {
...
         } else if (associatedFile) {
             FileAssociationService.getInstance().handleFileAssociation(associatedFile);
+        } else if (process.platform === "linux") {
+            createWindow("index.html");
         } else {
-            createWindow(process.platform === "linux"
-                ? "index.html" : "launcher.html"
-            );
+            const autoUpdate = StaticConfigurationService.getInstance().get("auto-update");
+            createWindow(autoUpdate === undefined || autoUpdate === true
```
silentrald, line 137: @Zagrios can you change this to just `.get("auto-update", true);` I forgot to change this when I added the default value as the 2nd parameter.

Zagrios: Done 👌

Zagrios:
@silentrald I will delay this one as well.
I think the feature must include a way for users to know whether an update is available or not.

Something like this should do the job (paint design 😅):
image

We already have the icon for the title bar: `DownloadIcon`.
A modal might also work.

When retrieving the latest available update, "electron-updater" provides us with the version name and its URL, I believe.

silentrald:
@Zagrios Should be good to re-review, set the AutoUpdate to NEVER by default, although couldn't test it on a windows. Might need to change the styling a little bit if needed.

Although not sure if we also need a "Don't remind me" button on the popover so it won't popup on start of the app if the app is outdated.

Zagrios:
- Done some fixes
- Changed the UI a bit:
Capture d'écran 2025-06-11 135429
- Enabled auto-update by default

Thanks @silentrald ❤️ 
Merging 🚀

Zagrios (approved):
Looks good to me 🙌

PR #790, opened 2025-02-07
[feat] Add build_start script for easier testing

TheUnifox:
## Scope

It's just easier to test when rebuilding is needed to have one command for build and start. It literally just calls build and start one after the other. Often a rebuild isn't needed when making changes, but when it is, having the one command to quickly build and start can be nice.

## Implementation

Just adding `"build_start": "npm run build && npm run start",` to `scripts` in `package.json`

## How to Test

- Run `npm run build_start`
- Wait for bs-manager to build and start

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "main": "./.erb/dll/main.bundle.dev.js",
     "version": "1.5.2",
     "scripts": {
+        "build_start": "npm run build && npm run start",
         "build-rust-scripts": "ts-node ./.erb/scripts/build-rust-scripts.js",
         "build": "concurrently \"npm run build:main\" \"npm run build:renderer\"",
         "build:dll": "cross-env NODE_ENV=development webpack --config ./.erb/configs/webpack.config.renderer.dev.dll.ts",


silentrald:
@TheUnifox just a question, but why do you need to build prod stuff and run the dev at the same time? Am I missing something here?

TheUnifox:
Do you not need to build before start? Am *I* missing something?

TheUnifox:
Probably would be good to add something saying *how* to test in `CONTRIBUTING.md`, instead of just saying "test". I thought I had to build and start, since nothing said anything for testing

silentrald:
@TheUnifox yeah, normally when I dev here, I just do

```bash
npm run start
```

For testing/publishing the checks are currently

```bash
npm run lint -- --fix
npm run test:unit
```

If you want to get an exe build, its

```bash
npm run package
```

Its different in linux as well but it should be the same.

silentrald:
The test scripts were recently added though as well. So yeah, might need to update the `CONTRIBUTIONS.md` file for how things work for developing. I'll ping up our docs guy.

PR #793, opened 2025-02-08
[bugfix] Don't fallback to system WINEPREFIX

silentrald:
Issue described in [discord](https://discord.com/channels/1049624409276694588/1337228531150749799)

Hide the `BSManager/SharedContent/compatdata` folder **before** accessing the mods tab.
image

Hide the `BSManager/SharedContent/compatdata` folder **after** accessing the mods tab.
Screenshot_20250208_093711

PR #795, opened 2025-02-08
[bugfix] show error when BSManager can't connect to beatmods

silentrald:
Blocked by  #793

Show different error when beatmods is unreachable or down.

image


GaetanGrd:
Hey, thanks for this addition. Would it be possible to have buttons instead of hyperlinks? (I find it more pleasant x) )

like : 
image
(Best Photoshopper, I don’t want to hear anything else XDD)

silentrald:
@GaetanGrd might be better for another PR though. Although I can't handle it this week since I am busy might let @Zagrios handle it if he wants to.

src/shared/constants.ts
```diff
@@ -0,0 +1,4 @@
+
+export const DISCORD_URL = "https://discord.gg/uSqbHVpKdV";
```
Zagrios, line 2: Thanks for that 🙏

Zagrios (approved):
Looks good to me 🙌

Zagrios:
Thanks @silentrald ❤️ 
Merging 🎉

PR #805, opened 2025-02-14
[bugfix] fix issue with extracting root map files

silentrald:
The regex for root map folders was broken, just extract the full zip instead.


Zagrios:
Thanks @silentrald ! ❤️ 
Merging 🚀

Issue #806, opened 2025-02-14
[BUG] : UUID folders not being removed after linking folders


LiamillionSS:
### Issue encountered

![Image](https://github.com/user-attachments/assets/a48442fa-e8ce-4a1e-90d0-8b3e9a28a757)

### Expected Behavior

Folders should be deleted afterwards

### Steps to Reproduce

1. Link folders
2. Observe

### Operating System

Linux

### Version

1.5.2

silentrald:
#807 is merged, closing this ticket. Thank you @LiamillionSS ٩(^ᗜ^ )و ´-

PR #807, opened 2025-02-14
bugfix/806 fix linux UUID folders not deleting

LiamillionSS:
## Scope

Closes ticket #806 

## Implementation

Added missing rmSync so UUID linking folders are deleted as expected


src/main/helpers/fs.helpers.ts
```diff
@@ -63,14 +63,15 @@ export async function deleteFolder(folderPath: string, options?: RmOptions) {
     }
 }
 
-export function deleteFolderSync(folderPath: string, options?: RmOptions) {
+export async function deleteFolderSync(folderPath: string, options?: RmOptions) {
```
Zagrios, line 66: @LiamillionSS @LiamillionSS You should not need to make this function async, since there is no `await` 👀

Zagrios (changes requested):

src/main/helpers/fs.helpers.ts
LiamillionSS: Fixed

Zagrios (approved):
Looks good to me 🙌

Zagrios:
Thanks @LiamillionSS ❤️ 
Merging 🚀

PR #838, opened 2025-04-01
Fix SteamVR opening in FPFC

Arimodu:
- Wait for the game to either close (keep it refed) or the user closes the app (unref and restore)

## Scope

A mostly small change to how starting the game is handled. Instead of unref'ing on a timeout, the game is kept ref'ed for the duration of its runtime, or until the launcher is closed by the user. 

## Implementation

Since this relates only to steam launches, I copied the `launchBs` implementation to the `steam-launcher.service`. The original implementation is still intact for other services.
Removed the timeout and instead now bind an event handler to will-quit electron event.
Promise can now resolve in three ways:
- Game fails to start -> original behavior
- Game starts, runs for any amount of time, then exits (cleanly or crash) -> Promise does not resolve until the game exits
- Launcher closes -> Game is unref'ed and steamVR is restored.

### Note
Not exactly sure if the promise should be resolved during the app exit. It could lead to a race condition where the launcher tries to restore the already restored SteamVR, however I dont think it matters much, since its output is catched anyway, so it shouldnt cause any issues

## How to Test
- Start the game in FPFC
- Check the SteamVR folder name -> it should be SteamVR.bak
- Close game or launcher -> SteamVR restored to original name

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/src/main/services/bs-launcher/steam-launcher.service.ts
+++ b/src/main/services/bs-launcher/steam-launcher.service.ts
@@ -12,6 +12,8 @@ import { UtilsService } from "../utils.service";
 import { exec } from "child_process";
 import { LaunchMods } from "shared/models/bs-launch/launch-option.interface";
 import { app } from "electron";
+import { ChildProcessWithoutNullStreams } from "child_process";
+import { SpawnBsProcessOptions } from "./abstract-launcher.service";
 
 export class SteamLauncherService extends AbstractLauncherService implements StoreLauncherInterface{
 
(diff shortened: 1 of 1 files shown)


Zagrios:
@Arimodu What about just passing the unrefAfter option to the launchBs function without changing anything else? 🤔

Something like this:
`steam-launcher.service.ts`
```ts
...
const spawnOpts: SpawnBsProcessOptions = { env, cwd: bsFolderPath, unrefAfter: minToMs(5) };

const launchPromise = !launchOptions.admin ? (
    this.launchBs(bsExePath, launchArgs, {
        ...spawnOpts,
        protonPrefix
    }).exit
)
...
```

If unrefAfter is not defined, the BS process is unrefed after 10 seconds, so setting it to 5 or 10 minutes might be enough 🤔

Arimodu:
I mean... fair I guess. If you put it in minutes theres no issue. 
however beatsaber stays refed in that case and some people close BSManager after the game starts. In that case afaik the game would get killed. My solution handles that by unrefing on app close OR after game closes, so no need to find a balanced delay

Arimodu:
Plus this opens up options for something along the lines of steams Stop button when the game is running, since BSManager is aware of the game process

Zagrios:
Yeah ok fair, that sounds good to me 👌

Zagrios:
Merged 🚀 
Thanks @Arimodu ! ❤️

PR #848, opened 2025-04-11
[bugfix] add system env to beat saber exe launch command for oculus

silentrald:
Issue described in [Discord](https://discord.com/channels/1049624409276694588/1359771444141097081)


Zagrios (approved):
Yeah, when I saw that process.env had been removed, I was wondering if it would break something and apparently it did 😅 Thanks for the quick patch 🙌

Zagrios:
Thanks @silentrald ! ❤️ 
Merging 🚀

Zagrios:
> Yeah, when I saw that process.env had been removed, I was wondering if it would break something and apparently it did 😅 Thanks for the quick patch 🙌

Oh `process.env` was only for Steam, i beleave, so forget about that 👌

PR #868, opened 2025-05-06
[bugfix-865] unref process to close window

### LINKED ISSUE #865: [BUG] : Using BSmanager Game Launch Shortcut will not close the BSmanager launching game window anymore.
### Issue encountered

The window that a game version is launching is not auto losing anymore. that is sadge because i use a capture card which means everything that is infront of the actual game is in my stream.
Did add a screenshot of the window.
I am on windows 11

Image

### Expected Behavior

After a big update a while ago the shortcut windows for launching a bsmanager game installation did close automatically after the game launched.

### Steps to Reproduce

_No response_

### Operating System

Windows 11

### Version

v 1.5.3

### Additional Context

_No response_

silentrald:
Closes #865 
Issue reported in [discord](https://discord.com/channels/1049624409276694588/1368882495298338876)

Was missed in #838, just added a timeout to unref the process to close the window.

### DIFF
--- a/src/main/services/bs-launcher/steam-launcher.service.ts
+++ b/src/main/services/bs-launcher/steam-launcher.service.ts
@@ -12,6 +12,7 @@ import { UtilsService } from "../utils.service";
 import { exec, ChildProcessWithoutNullStreams } from "child_process";
 import { LaunchMods } from "shared/models/bs-launch/launch-option.interface";
 import { app, Event } from "electron";
+import { sToMs } from "shared/helpers/time.helpers";
 
 export class SteamLauncherService extends AbstractLauncherService implements StoreLauncherInterface{
 
@@ -67,6 +68,8 @@ export class SteamLauncherService extends AbstractLauncherService implements Sto
     protected launchBs(bsExePath: string, args: string[], options?: SpawnBsProcessOptions): {process: ChildProcessWithoutNullStreams, exit: Promise<number>} {
         const process = this.launchBSProcess(bsExePath, args, options);
 
+        let timeoutId: NodeJS.Timeout;
+
         const exit = new Promise<number>((resolve, reject) => {
             // Don't remove, useful for debugging!
             // process.stdout.on("data", (data) => {
@@ -101,6 +104,16 @@ export class SteamLauncherService extends AbstractLauncherService implements Sto
             });
 
             app.on('will-quit', onWillQuitHandler);
+
+            const unrefAfter = options?.unrefAfter ?? sToMs(10);
+            timeoutId = setTimeout(() => {
+                log.error("BS process unref after timeout", unrefAfter);
+                process.unref();
+                process.removeAllListeners();
+                resolve(-1);
+            }, unrefAfter);
+        }).finally(() => {
+            clearTimeout(timeoutId);
         });
 
         return { process, exit };


Arimodu:
Oh... seems like I made a mistake when changing that, tho reimplementing the timeout would make the point of my PR and change moot. 

The thing I was trying to solve was that for some people with either slower PCs (like my laptop) or a lot of mods, steamvr is restored before the game tries to launch openxr, causing steamvr to open. 

I will look into closing the window itself on a timeout instead of unrefing the BS process (tho considering that it prolly closes BSManager id say it will unref anyway). I dont use that feature so I didnt notice that issue

Arimodu:
Okay, I tested and implemented locally the following instead.
I dont want to step over your PR so ill post this here instead of pushing up to GH:
```js
...
            app.on("will-quit", onWillQuitHandler);

            // This should be fine, the above callbacks are still bound so they should still run regardless.
            promiseTimeoutId = setTimeout(() => {
                log.info("Resolved promise after timeout (20s)");
                resolve(-1); // Put this in a timeout for visual effect (game window may take a bit to open, lets trick users into thinking stuffs happening...)
            }, sToMs(20));
        }).finally(() => {
            clearTimeout(promiseTimeoutId);
        });

        return { process, exit };
    }
```

silentrald:
@Arimodu Is there any reason to still attach the Beat Saber process with BSM after the timeout?

> Forgot to mention that for this to work as intended, I moved the restore steamvr into the callbacks instead of the finally block in launch function

Also, can you push your fix onto your fork? I'll try to see what changes you did as well everything is good.

Arimodu:
Here is what I tested: https://github.com/Arimodu/bs-manager/commit/190fd0a875e2bb24346be905452ed9927e6c4b50

As for the reason... I explained that in my PR, but basically on my laptop for example, when I start with FPFC SteamVR launches 100% of the time because the game takes a good minute to actually start. I personally dont use the shortcuts, in fact I didnt even know that was a feature until I saw this PR. 

Staying attached to the process is probably the best way to keep steamvr out of the picture without having to do like some shenanigans with windows services and whatnot, as in this case, the only times steamvr is renamed back is if the game crashes, exits, or the launcher exits. In other words, fixes the issue for basically everyone except ppl who use the shortcuts (where the launcher closes after the promise resolves). 

To solve this even for people who do use the shortcuts, well... because BSIPA Log is not emmited to STDOUT my best guess is to monitor _latest.log for changes and wait for some menu-only line, or to inject some sort of helper mod into the game which just zenjects onMenu and signals BSM to exit. The helper mod is the better of the two, and id love to implement it but I dont think it would get merged. 

Also as I did mention in my original PR, keeping the game refed opens BSM to something akin to the Steam "Stop" button

silentrald:
@Arimodu I think creating a mod is the way to go since I had a conversation with someone with this that we can get an accurate startup of the game then we can unref and/or close BSM. (Had an idea for a feature where closing BSM after a proper launch occurred).

Although the only problem is that its a mod, if its a BSIPA dependency, then unmodded games will still rely with timeouts.

Although your idea with Zenject, though not that familiar with it, if we can run the Beat Saber executable with some BSM dll to get a signal/trigger, I got no problem with it tbh, as long as its not dependent nor interfere with BSIPA.

@Zagrios any thoughts on this as well, I might get tunnelled vision somewhere?

Arimodu:
> as long as its not dependent nor interfere with BSIPA.

Thats more than a tall order there. You are talking about injecting directly from BSM, which would need some native code, or powershell hooks or something along those lines. I am not extremely familiar with injecting dlls into processes but from what I know, you need native APIs for that (`CreateRemoteThread` or `LoadLibrary` on windows id guess, no idea with linux). 

When I wrote the above, I was thinking with BSIPA and Sirautil as a dependencies. An unmodded game wont have issues with load times even in the worst of circumstances, the issue arises with modding anyway, so id keep the timeout for unmodded games, and just use a named pipe or socket which a modded game would send scene name updates to. 

Note that my original idea was to depend on sirautil to use zenject, because it would make the code simpler and smaller + you get the App, Menu, GameCore, etc... hooks to which you can inject classes. [Here](https://github.com/Arimodu/CamControl/blob/master/CamControl/Plugin.cs) is a mod that a friend comissioned that uses this. The two dependencies are IPA and SiraUtil + Cam2 because its purpose is to control cameras.

I am not saying its not possible to do this dependency-less. Its just there is so much complexity that I am not exactly sure if the gains from it would be worth the effort.

As for if its required to use zenject... probably not. You could just go straight to `SceneManager.sceneLoaded` and get your current scene name that way. In fact thinking about it now, using zenject to do this might be way overkill, since `SceneManager` has everything we need and its static, so no instance needed.

I could whip up something this afternoon to test, but me being clumsy with JS, its not gonna be pretty on the BSM side. 

If you want to chat about this more in-depth, you can find me on BSMG discord under the same name, shoot me a msg or something :)

silentrald:
@Arimodu that might work then. So probably the implementation should go like this then.

1. Have a check whether a BS version is modded or not (should contain a BSIPA)
2. Create the mod for unrefing, would probably just do stdout and pattern matching unless there's another elegant way of doing this.
3. Hook it up with the launch:
* Unmodded - Using timeouts
* Modded - Using the created mod

I'll double check this on linux if it works as well, ps. linux just has a windows wrapper so technically its just the same as running in windows.

Tell me if I've missed something. Also might delegate the task to you since you have exp with creating mods, might close the PR after you reply/emote to this.

Arimodu:
Alright, sounds good. 

Might want to keep this PR open till @Zagrios sees it tho. Or we might want to make a group DM on this or something...

PR #870, opened 2025-05-09
[bugfix] Fix mod installation on NixOS

ImSapphire:
## Scope

Inherits environment when starting wine for BSIPA installation, which matches behaviour when launching Beat Saber.
Use selected Proton's wine instead of system wine, since the wine prefix version may mismatch and cause wine to crash.

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/src/main/services/linux.service.ts
+++ b/src/main/services/linux.service.ts
@@ -129,12 +129,7 @@ export class LinuxService {
         return fs.pathExistsSync(protonPath) && fs.pathExistsSync(winePath);
     }
 
-    public async getWinePath(): Promise<string> {
-        if (await this.isNixOS()) {
-            // Use system wine for nixos
-            return "wine";
-        }
-
+    public getWinePath(): string {
         if (!this.staticConfig.has("proton-folder")) {
             throw new Error("proton-folder variable not set");
         }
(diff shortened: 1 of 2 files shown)


silentrald:
@ImSapphire I might double check on the flatpak build if I can install mods with this. There are some weird env vars that get injected there that I assume might break flatpak.

@Mistyttm can you help me verify this if everything's working as expected since your on NixOS.

Mistyttm:
@silentrald can already confirm it is my friend, these patches were on the nixos package before Sapphire brought them here. They managed to get it all working

silentrald (approved):
Should be good to merge (ﾉ^ヮ^)ﾉ*

@ImSapphire would this needs to be merged first or the other PR from NixOS? Or they are independent to one another.

cc @Mistyttm @Zagrios

Mistyttm:
@silentrald They are independent as of now. We're pulling from the tagged 1.5.3 release, so when the tagged 1.5.4 release with this code comes out, we can update the package on nixpkgs and remove the patches on our end

nixpkgs version is completely independent from this :)

Zagrios:
Merging! 🚀 
(Sorry for the delay 😅)

PR #872, opened 2025-05-16
[fix] use a caching method for ipv4/6 request calls

silentrald:
~~Somewhat support ipv4, unlike the original implementation was to always check between the 2, let the user either use ipv6 (default) or force ipv4 in the settings.~~

Applied logic from https://github.com/Zagrios/bs-manager/pull/872#issuecomment-2954092281 insteadd

### DIFF
--- a/src/main/services/mods/beat-mods-api.service.ts
+++ b/src/main/services/mods/beat-mods-api.service.ts
@@ -76,8 +76,7 @@ export class BeatModsApiService {
         }
 
         return this.requestService.getJSON<{ modVersions: BbmModVersion[] }>(
-            `${this.MODS_REPO_API_URL}/hashlookup?hash=${hash}`,
-            { silentError: true }
+            `${this.MODS_REPO_API_URL}/hashlookup?hash=${hash}`
         ).then(({ data }) => {
             this.updateModsHashCache(data?.modVersions ?? []);
             return data?.modVersions?.at(0);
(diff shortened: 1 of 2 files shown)


Zagrios:
Hmmm, users shouldn't have to touch that 🤔 It should just work.
What about restoring the check between IPv4 and IPv6, but doing it better?
Like:
- Using a cache for the working IP version per website, something like:
```json
{
    "beatsaver.com": 4,
    "beatmods.com": 6,
    ...
}
```
- Filtering request errors, for example, don’t retry if the website was reached but returned an actual error
What do you think?

I’m also okay with merging this PR for now and improving that later 👌

silentrald:
@Zagrios Instead of a request check, why not use something like a `dig` command so that we can know if the ip address that we can get is either a ipv4/6.

We can still use the caching above as well, only thing that might be missed if it fails. Do we just have a clear cache as well, or do the fallback? ie.

- User request website with family found in cache but fails
- User will do the "dig" command
  - If family is the same, this is an error
  - Else, proceed with another request
    - Success, write over the cache then proceed
    - Failure, this is an error
Dig: Can refer to this method, should be within the nodejs package https://nodejs.org/api/dns.html#dnslookuphostname-options-callback

Zagrios (approved):
Looks good to me 🙌

Zagrios:
Thanks @silentrald ❤️ 
Merging 🚀 

Hopefully the cache is enough, we'll see 😅 Worst case, we'll improve it somehow, but I think we'll stick with this logic 👌

PR #873, opened 2025-05-16
[fix] return null if downloadMap will incur an error

silentrald:
Issue described in [discord](https://discord.com/channels/1049624409276694588/1372636338913939497)

`downloadMap` throws an error instead of returning a null value, created a try catch wrapper to avoid the issue.

### DIFF
--- a/src/main/services/additional-content/maps/local-maps-manager.service.ts
+++ b/src/main/services/additional-content/maps/local-maps-manager.service.ts
@@ -448,18 +448,23 @@ export class LocalMapsManagerService {
             return installedMap;
         }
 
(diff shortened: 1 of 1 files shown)


Zagrios:
@silentrald Errors should be handled by whoever calls this function. In most cases, it's not an issue since errors are passed back to the frontend through the `ipc-service` (and usually a custom error message is displayed).
But it's in `downloadPlaylistSongs` that it becomes a problem.

I think just adding a continue; here when an error occurs should be enough (a bit like with mapDetail).
image

Zagrios:
Thanks @silentrald ! ❤️ 
Merging 🚀

PR #889, opened 2025-06-21
[fix] fixed issue with double http prepending on http proxy

silentrald:
http gets prepended even if the url from the regedit entry already has it


Zagrios (approved):
Looks good to me 🙌

Zagrios:
Thanks @silentrald ❤️ 
Merging 🚀

PR #894, opened 2025-07-03
[bugfix] prevent corruption of the game when installing BSIPA

Meivyn:
There's a reason why Mod Assistant and other installers are looking for the presence of `winhttp.dll` before executing IPA.exe. While I despise this executable and plan to remove it in a future update, if it is launched right after a game update and before the user launches the game, it will restore the backup of the previous game version and break the game as a result.

It is automatically run when needed by BSIPA itself and should not be run manually unless instructed otherwise.

There's no way currently to prevent this from happening when uninstalling BSIPA.

### DIFF
--- a/src/main/services/mods/bs-mods-manager.service.ts
+++ b/src/main/services/mods/bs-mods-manager.service.ts
@@ -49,12 +49,11 @@ export class BsModsManagerService {
                 return undefined;
             });
 
-        if(mod?.contentHashes?.some(content => content.path.includes("IPA.exe"))){
+        if (mod?.contentHashes?.some(content => content.path.includes("IPA.exe"))) {
             return undefined;
         }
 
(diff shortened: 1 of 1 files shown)


src/main/services/mods/bs-mods-manager.service.ts
```diff
@@ -207,13 +206,11 @@ export class BsModsManagerService {
             processIPA.once("exit", code => {
                 clearTimeout(timeout);
                 if (code === 0) {
-                    log.info("Ipa process exist with code 0");
```
Meivyn, line 210: For the record, I don't think this is ever necessary as it would time out, but feel free to change this if I'm wrong.

Zagrios:
Thanks @Meivyn ❤️ 
Merging 🚀

PR #895, opened 2025-07-08
[fix] handle launch option string properly

silentrald:
Issue described in [discord](https://discord.com/channels/1049624409276694588/1391537835181605006)

Still working on
- [ ] Remove on `parseEnvString` and just return the list that was parsed properly.
- [ ] Add test cases for `parseLaunchOptions` and `parseEnvString` to handle the commands described in the discord post.


silentrald:
@Zagrios can you check if this doesn't break on windows, also getting this checked by a user in linux as well if it doesn't break on their end.

Zagrios:
Thanks @silentrald ! ❤️ 
Merging 🚀

PR #912, opened 2025-08-29
[bugfix] 1.40.10 seen as 1.40.1

Zagrios:


f1amy:
@Zagrios on BSM v1.5.4 after launch I had 1.40.10 and 1.40.9 installed
Then I removed 1.40.9 and 1.40.10 changed to 1.40.1
After relaunching the app its detected as 1.40.10 as normal

Seems like this fix is needed copied somewhere else

PR #918, opened 2025-08-31
Add workaround for headset with canted views

tleydxdy:
See https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2312

## Scope

Reflections in beat saber can look wrong when running with monado + openvr translator

## Implementation

I tried to use the launch options menu to customize this at first but couldn't get it to work. Since OXR_PARALLEL_VIEWS only affect canted views headsets and only recognized by Monado, it should be safe to set by default and not affect other headset/setups.

## How to Test

Need a headset with canted views like the Index and run with monado, reflections should look cross-eyed by default and with this change it should look correct.

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/src/__tests__/unit/os.test.ts
+++ b/src/__tests__/unit/os.test.ts
@@ -51,6 +51,7 @@ describe("Test os.helpers bsmSpawn", () => {
                 STEAM_COMPAT_CLIENT_INSTALL_PATH: "/steam",
                 STEAM_COMPAT_APP_ID: BS_APP_ID,
                 SteamEnv: "1",
+                OXR_PARALLEL_VIEWS: "1",
             });
         }
     });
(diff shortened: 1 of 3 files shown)


silentrald:
@tleydxdy Is it possible to create a launch options button for this instead, rather than setting it defaultly to all linux systems (since not all uses monado as their runtime). Would probably suggest to add it as a default button here (the button label is like "Monado" or something specific to the OXR_PARALLEL_VIEWS env var).

image

tleydxdy:
Yeah I can try that, although since this envvar only have an effect on monado, I don't think it would be a problem to just always set it so it works ootb.

silentrald:
@tleydxdy If thats the case, no issue. Just a minor comment on `abstract-launcher.service.ts` to add a comment that the env var is monado specific as well and Im all goods with the PR.

Zagrios:
Thanks a lot @tleydxdy ❤️ 
Merging ! 🚀

Issue #922, opened 2025-09-01
[BUG] : Unable to get original Steam version on Linux


Corben78:
### Issue encountered

Version 1.5.4 introduces the following error:

```
[1] 11:14:31.794 › Unable to get original Steam version Error: No Beat Saber Steam version found
[1]     at BSLocalVersionService.getSteamVersion (/home/corben78/git/bs-manager/.erb/dll/main.bundle.dev.js:147610:19)
[1]     at async BSLocalVersionService.getInstalledVersions (/home/corben78/git/bs-manager/.erb/dll/main.bundle.dev.js:147626:30)
```

PR https://github.com/Zagrios/bs-manager/pull/907 changes the `getPath("home")` to the data home path, which is not the home directory of the user anymore.
The steam default directories are either `$HOME/.steam/steam` or `$HOME/.local/share/Steam`.

The current implementation of `getSteamPath()` results in `/home/corben78/.local/share/.steam/steam` which is invalid.

I don't think it's a good idea to have the `getPath("home")` function point to a subdirectory of the home directory, it should return the home directory directly. Especially as `getPath("home")` is used in other locations as well.

As `$HOME/.steam/steam` is valid to get the steam path, I don't think a change of `getPath("home")` necessary at all. Or `getSteamPath()` should be adjusted not to use `getPath("home")` but e.g. `getPath("data_home")`.

### Expected Behavior

Steam Beat Saber installation is properly located

### Steps to Reproduce

Install 1.5.4, start bs-manager

### Operating System

Linux

### Version

1.5.4

Corben78:
As I see the original purpose of this PR, maybe it's better to have the installation path in its own variable instead of re-using and altering the $HOME path.

c-x-berger:
I worked around this in the stupidest way possible - I made a symlink at `$HOME/.local/share/.steam` to `$HOME/.steam`. BSManager _seems_ to work OK - mods install etc, but this seems like the sort of thing that will come back to bite me later.

Issue #906, opened 2025-08-13
[FEAT.] [Linux]: Change default instance installation folder to "$XDG_DATA_HOME" and/or "$HOME/.local/share"


WearyOlly:
### Problem

I find that BSManager defaulting the instance installation to $HOME on first bootup quite annoying, since it then clutters up my home folder

### Solution

Changing the default installation folder to $XDG_DATA_HOME and/or $HOME/.local/share would do 2 things:
1. clean up the home folder,
2. makes BSManager XDG Base Directory Specification compliant

### Operating System

Linux

### Additional Context

XDG Base Directory Specification: https://specifications.freedesktop.org/basedir-spec/latest/

PR #923, opened 2025-09-01
[bugfix] Adjust steam path on Linux (#922)

Corben78:
[bugfix] Adjust steam path after changing getPath("home") to $XDG_DATA_HOME (#922)

## Scope

[closes TICKET-922](https://github.com/Zagrios/bs-manager/issues/922)

## Implementation

Use `$HOME/.local/share/Steam` as location. `getPath("home")` results in `$HOME/.local/share` already, so just "Steam" needs to be added (capital).

### DIFF
--- a/src/main/services/steam.service.ts
+++ b/src/main/services/steam.service.ts
@@ -76,7 +76,7 @@ export class SteamService {
 
         switch (process.platform) {
             case "linux":
-                this.steamPath = path.join(app.getPath("home"), ".steam", "steam");
+                this.steamPath = path.join(app.getPath("home"), "Steam");
                 return this.steamPath;
             case "win32": {
                 const res = await list(["HKLM\\SOFTWARE\\Valve\\Steam", "HKLM\\SOFTWARE\\WOW6432Node\\Valve\\Steam"]);


Corben78:
As #906 wants the default install path on Linux changed, while getPath("home") is used as a location, it seems easier to adjust the `getSteamPath()` function for now.
`getPath("home")` is also used on other locations, so I'm not sure which other side effects introducing a new path name like `getPath("install")` would have.

silentrald:
@Corben78 hmmmm yeah probably the safest to introduce the `app.getPath("install")` in that case since in the future, I might need to getPath("home") tbh. Note that you should also add it to windows as well ie. 
```js
app.setPath("install", app.getPath("home"))
```

Corben78:
I'll have a look, I don't use Windows much, so I need a bit longer for testing.
For now I'd suggest this PR just to fix the current issue and then have a refactor addressing the install path.

silentrald:
@Corben78 yeah fix should be fine. Still contemplating since `.steam/steam` is backward compatible but will accept this for now and just wait for users if they report that the Steam folder is incorrect.

@Zagrios any comment or this is good?

In any case, will try to add the refactor of this code when I'm available. Hopefully by next week.

Zagrios:
Fixed in #943 (thanks to @silentrald)

PR #948, opened 2025-10-12
fix bsipa installation error on linux #945

### LINKED ISSUE #945: [BUG] : BSIPA: "Game does not seem to be a Unity project. Could not find the libraries to patch."
### Issue encountered

When trying to install BSIPA via the mod manager under linux, it errors out with BSIPA saying `ERROR: Game does not seem to be a Unity project. Could not find the libraries to patch.`

### Expected Behavior

_No response_

### Steps to Reproduce

1. Open a game version with the core mod `BSIPA` not yet installed.
2. Open mods tab.
3. Press `Install or update`.
4. Observe error message.

Image

### Operating System

Linux

### Version

1.5.4.r18.g56f80f1

### Additional Context

The logs state that it ran the command:

```bash
"/home/withering/.local/share/Steam/steamapps/common/Proton - Experimental/files/bin/wine64" "/home/withering/.local/share/BSManager/BSInstances/1.40.8/IPA.exe" "/home/withering/.local/share/BSManager/BSInstances/1.40.8/Beat Saber.exe" -n
```

formatted for viewing pleasure:

```bash
proton="$HOME/.local/share/Steam/steamapps/common/Proton - Experimental/files/bin/wine64"
share="$HOME/.local/share/BSManager"
version="1.40.8"

"$proton" "$share/BSInstances/$version/IPA.exe" "$share/BSInstances/$version/Beat Saber.exe" -n
```

When I run this command in terminal I can replicate the exact error, what's weird is when I `cd` into `$share/BSInstances/$version/` and then run:

```bash
"/home/withering/.local/share/Steam/steamapps/common/Proton - Experimental/files/bin/wine64" "/home/withering/.local/share/BSManager/BSInstances/1.40.8/IPA.exe" "Beat Saber.exe" -n
```

formatted:

```bash
proton="$HOME/.local/share/Steam/steamapps/common/Proton - Experimental/files/bin/wine64"
share="$HOME/.local/share/BSManager"
version="1.40.8"

cd "$share/BSInstances/$version"

"$proton" "$share/BSInstances/$version/IPA.exe" "Beat Saber.exe" -n
```

Then it appears to work fine, **BUT NOT** if I `cd` into it and use the same absolute path, only if its a relative path.

notwithering:
Previously, attempting to install the BSIPA core mod would cause it to fail with the error "Game does not seem to be a Unity project. Could not find the libraries to patch." This is because the env was being overwritten after applying the WINEPREFIX, this causes BSIPA installation fail because winepath will use the drives configured in whatever the default prefix is, my Linux root was set to `U:` which is different to `Z:` used by the proper prefix. 

This PR initializes the env with the process.env and then writes the prefix. This closes #945

image

## Test

1. Be on Linux using any version of Proton
2. Open a installation with BSIPA not yet installed. 
3. Press install button.
4. Observe that it successfully installs now.

I have not tested it on Windows yet so if someone could do that before approving that would be greatly appreciated, I don't see why it wouldn't work though since the changed code is basically all under a Linux check.

### DIFF
--- a/src/main/services/mods/bs-mods-manager.service.ts
+++ b/src/main/services/mods/bs-mods-manager.service.ts
@@ -156,7 +156,7 @@ export class BsModsManagerService {
             return false;
         }
 
-        const env: Record<string, string> = {};
+        const env: Record<string, string> = { ...process.env };
         const cmd = `"${ipaPath}" "${bsExePath}" ${args.join(" ")}`;
         let winePath: string = "";
         if (process.platform === "linux") {
@@ -176,7 +176,6 @@ export class BsModsManagerService {
                 throw new CustomError("Could not find BSManager WINEPREFIX path", "no-wineprefix");
             }
             env.WINEPREFIX = winePrefix;
-            Object.assign(env, process.env);
         }
 
         return new Promise<boolean>(resolve => {


notwithering:
Also I had a decent amount of trouble getting bs-manager up and running, maybe consider better tooling, more guidance in the contributing guide, or some shell scripts to help anyone new? I also had troubles with windows specific packages since I'm on linux that I had to do some ebr configuration with to ignore.

notwithering:
Made some edits to the PR, now explains why exactly the previous implementation didn't work, and why this one works.

Also sorry for the messy commit history

silentrald:
@Zagrios tested on my environment and it seems to be working as expected. Could you check in windows if this will not be an issue.

Thanks @notwithering (ﾉ^ヮ^)ﾉ*.

notwithering:
Also it should be noted that just using "Beat Saber.exe" only works because it is both a valid Unix and windows style path, if you need more resistance against regressions, I can implement a way to convert the absolute Unix path into a windows path, that will make it so if you ever change BS_EXECUTABLE to "randomdir/my.exe" or something it will still work

silentrald:
@notwithering shouldnt the pathing be translated by wine anyway? Since I did check that previously when I checked under the hood on why pathing works, look into `winepath` since thats the pne that should translate paths from unix to windows style pathing and why isnt working in your system.

notwithering:
I'm not going to lie, this has been my most inefficient debugging process, but I think I finally found the real issue.

in the `executeIPA` function in `bs-mods-manager.service.ts`

```typescript
    private async executeIPA(version: BSVersion, args: string[]): Promise<boolean> {
...
        const env: Record<string, string> = {};
        const cmd = `"${ipaPath}" "${bsExePath}" ${args.join(" ")}`;
        let winePath: string = "";
        if (process.platform === "linux") {
...
            env.WINEPREFIX = winePrefix;
            Object.assign(env, process.env);
        }
```

look closer at

```ts
            env.WINEPREFIX = winePrefix;
            Object.assign(env, process.env);
```

that is overwriting the assignment with the current env, an easy solution is just swapping those lines or what I prefer:

```ts
    private async executeIPA(version: BSVersion, args: string[]): Promise<boolean> {
...
        const env: Record<string, string> = { ...process.env };
        const cmd = `"${ipaPath}" "${bsExePath}" ${args.join(" ")}`;
        let winePath: string = "";
        if (process.platform === "linux") {
...
            env.WINEPREFIX = winePrefix;
        }
```

initializing the env with the process.env directly in there then the later assignment will stick

I've tested this hypothesis, and it is a solution to the problem. Now, the reason I think that nobody else has reported this problem is because

1. didn't notice the WINEPREFIX was wrong
2. the letter of their drive on their default wine prefix corresponding to their linux root was `Z:` which matches the default drive letter on the BSManager prefix

The only reason this affected me especially is because the drive letter corresponding to my linux root for my default WINEPREFIX is `U:` not `Z:` which caused a mismatch in what `winepath` reported but that the prefix expected

notwithering:
@silentrald Shall I restore and push the new changes?

silentrald:
@notwithering so does that mean your winepath is reporting U and not Z? If thats the case, we might need to run the wineprefix under the winepath then add it to the env var or is there an pathing format that would satisfy the path correctly without ising winepath.

notwithering:
@silentrald No, BSManager is writing the correct prefix, `env.WINEPREFIX = winePrefix;` but the next line is overwriting it with the current env, `Object.assign(env, process.env);`, setting that to the default, I was saying that I believe the BSIPA error was especially an issue for me is because my personal wineprefix, `~/.wine`, is configured to have the linux root on drive `U:` while other people would most likely have the default of `Z:` which is the same as the BSManager prefix, therefor `winepath` would technically report the same `Z:\.local\share\BSManager\SharedContent\compatdata\pfx` on both prefixes if both had the same `Z:` drive, but since my personal prefix uses `U:` it was reporting `U:\.local\share\BSManager\SharedContent\compatdata\pfx` which is not valid on the BSManager prefix.

TLDR: BSManager sets the correct WINEPREFIX but later overwrites it, making the incorrect WINEPREFIX used in the executed command.

silentrald:
@notwithering so you just need to switch those 2 lines. You could check if that works, i never thought it would overwrite that.

If you know how to rebase, should be fine. Or recreate the branch with master then force push your changes. As long as your commit tree is clean, but we do squash them iirc.

notwithering:
I'm proud of myself, I was able to restore it without absolutely wrecking the commit history. :rofl:

notwithering:
image

silentrald:
@Zagrios should be good for linux, just double check in windows since the env vars are now reinjected again on the launch.

Thank you again @notwithering (⁠｢⁠`⁠･⁠ω⁠･⁠)⁠｢

Zagrios:
No problem on Windows @silentrald  👌 
Thanks @notwithering ❤️ 
Merging ! 🚀

PR #974, opened 2025-12-11
fix: Playlist search modal losing state

AngeloTadeucci:
## Scope

Fixed an issue where the playlist search state was reset when navigating back from playlist details view.

## Implementation

The problem was caused by React remounting modal components on every render due to unstable keys. The `Modal` component was using `crypto.randomUUID()` as the `key` prop for each modal on every render, causing React to treat modals as new components and lose their internal state.

**Changes made:**

1. **Added stable ID to `ModalObject` type** (`modale.service.ts`):
   - Added an `id: string` field to the `ModalObject` type definition

2. **Generate ID once per modal** (`modale.service.ts`):
   - Modified `openModal()` to generate a unique ID using `crypto.randomUUID()` when creating the modal object
   - This ID is now generated once and persists for the lifetime of the modal

3. **Use stable ID as React key** (`modal.component.tsx`):
   - Changed the modal rendering to use `modal.id` as the key instead of generating a new UUID on each render
   - Also changed the overlay key to a static string for consistency

This ensures that React correctly identifies modal components across re-renders, preserving their internal state (including search results, scroll position, and form data).

## Videos
# Before
https://github.com/user-attachments/assets/75327e44-5a54-42b6-bc91-b20654cc86a7

# After

https://github.com/user-attachments/assets/9ae0c047-3832-402e-94f4-77bc4526d914

## How to Test

1. Search for playlists using any search term
2. Click on any playlist to view its details
3. Close the playlist details modal
4. Verify that the search results are still displayed (previously they would be reset)

## Emoji Guide

> 🟢 Nice refactor!

### DIFF
--- a/src/renderer/components/modal/modal.component.tsx
+++ b/src/renderer/components/modal/modal.component.tsx
@@ -74,9 +74,9 @@ export function Modal() {
 
     return (
             <AnimatePresence>
-                {currentModal ? <motion.span key={crypto.randomUUID()} onClick={onOverlayClicked} className="fixed size-full bg-black z-[90]" initial={{ opacity: 0 }} animate={{ opacity: currentModal && 0.6 }} exit={{ opacity: 0 }} transition={{ duration: 0.2 }} /> : undefined}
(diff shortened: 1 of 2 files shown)


Zagrios (approved):
Looks good to me 🙌

Zagrios:
Thanks a lot @AngeloTadeucci ! ❤️ 
Merging 🚀

PR #943, opened 2025-10-07
[bugfix-922] fix linux steam pathing issue

silentrald:
Closes #922, #923

Should resolve the steam pathing issues with linux builds.

#923 would be obselete with this fix.

PR #984, opened 2026-01-06
Updated NixOS Troubleshooting section

Reenotar:
## Scope

Updating Wiki Documentation for NixOS 25.11 installation troubleshooting


silentrald:
@Reenotar Already applied the patch in #943. The PR is unneccesary in any case since this is flagged as a bug.

Reenotar:
@silentrald Okay, I thought it should've been mentioned at least _temporarily_, until the patch will be official, because it has been 3 months since it has been solved and I think I wasn't the only one, who would've appreciated that the _wiki_ would let me know about the temporal fix, that's still in the production. Thanks for the comment, sorry to bother you :)

PR #999, opened 2026-01-27
Add UTM parameters for mod repo website links

frto027:
Make it more friendly for data sources by letting them know that the user is coming from bsmanager.


frto027:
Maybe more parameters could be added, e.g. installed game versions or bsmanager versions.
It's not very important, but the data source can use these data to provide some user-friendly responses. It depends on you.

Zagrios:
Hi @frto027, I'm closing this PR since I reverted the previous one. 
We can't really add an external API that hasn't been verified by anyone too risky on our end. 
If you get your site reviewed and approved by BSMG first, we can definitely reconsider. 
Thanks anyway 😉

PR #1012, opened 2026-02-08
[fix]: make OXR_PARALLEL_VIEW=1 toggleable rather than default

silentrald:
Discussed in LVRA that this may have issues with bsb linux users. So instead making the OXR_PARALLEL_VIEW=1, add it as an launch option button.

~~**TODO: Translation**~~


Zagrios:
I updated the branch with the current master changes and resolved the Linux launch environment conflict so the new parallel-views option stays opt-in while the newer Monado texture workaround remains in place.

Zagrios:
I pushed a small follow-up that keeps the new Parallel Views option intact when creating launch shortcuts and adds tests for the Linux launch environment. This keeps the option default-off while making sure it survives shortcut flows when selected.

Zagrios:
Pushed Zagrios-authored remediation commit cedb4ee803c5304b8336afe1c4c4e8d5868e6ca1 for the phase-2 Codex findings: shortcut links now preserve the existing Map Editor launch mod, and Linux shortcut/flatpak regression coverage now verifies OXR_PARALLEL_VIEWS stays opt-in. Local verification passed for git diff --check, npm run build, npm run test:unit, and changed-file ESLint; npm run test/lint still hit the existing Windows/Git-Bash test-pattern and nested-worktree ESLint config issues documented in the Discord log.

Zagrios (approved):
Looks good to me, thanks @silentrald!

PR #1025, opened 2026-03-05
[fix]: fpfc launch delay

Umbranoxio:
## Scope

fpfc launch takes 60+ seconds when steamvr lives in a protected dir (annoyingly even happens in oculus mode on the steam version of the game)

this seems to have been around for awhile, unsure how it hasn't been reported; anyway.. I've recently simplified my pc (big nvme, one partition) so i'd never run into this issue before as stuff was usually installed outside of `C:\`

anywho bsmanager backs up steamvr (`SteamVR` -> `SteamVR.bak`) before every fpfc launch to prevent steamvr from opening alongside the game and runs even when oculus mode is enabled even though `-vrmode oculus` bypasses steamvr entirely, so that backup is pointless. if steam is installed somewhere like program files, the rename hangs for ~60 seconds waiting on an EPERM before failing and moving on.

## Implementation

- skip the steamvr backup entirely when oculus mode is enabled -- it's not needed
- for the non oculus fpfc case, wrap the rename in a 5s timeout so a permission failure doesn't stall the whole launch, and show a warning toast telling the user that running the program as admin would give them a much better time
- actually wired up the `BSLaunchWarning` events to actually show toast notifications, looks like they were silently filtered out before

## How to Test

- steam in a protected directory like `C:\Program Files (x86)\Steam`, bsmanager running without admin
- launch any version with fpfc + oculus mode enabled -- should launch near instantly
- launch with fpfc only (no oculus mode) -- should launch within a few seconds, admin warning toast should appear

### DIFF
--- a/assets/jsons/translations/de.json
+++ b/assets/jsons/translations/de.json
@@ -561,6 +561,16 @@
                 "actions": {
                     "STEAM_NOT_RUNNING": "Starte Steam"
                 }
+            },
+            "warnings": {
+                "titles": {
+                    "FPFC_NEED_ADMIN": "FPFC mode setup incomplete",
+                    "UNABLE_TO_LAUNCH_STEAM": "Could not launch Steam"
+                },
+                "msg": {
+                    "FPFC_NEED_ADMIN": "BSManager needs administrator privileges to properly set up FPFC mode. Run BSManager as administrator for the best experience.",
+                    "UNABLE_TO_LAUNCH_STEAM": "Steam could not be started automatically."
+                }
             }
         },
         "steam": {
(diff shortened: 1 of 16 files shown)


Umbranoxio:
Also the concept of the steamvr restore being reliant on bsmanager gives me the willies

If a user:
- Launches game in fpfc through bsmanager
- Closes bsmanager
- Closes game
- Their SteamVR has been yeeted

Might be worth checking if a restore is pending on close of bsmanager or better yet; spawning a separate daemon process

Zagrios (approved):
Looks good after checking the launch changes and CI.

PR #1064, opened 2026-07-08
Replace "Link all" by "Unlink all" when all folders are linked

### LINKED ISSUE #360: [FEAT.] : Replace "Link all" by "Unlink all"
Currently, when you click on "link all" in "shared folder", you can't unlink everything at the same time... it's quite annoying, I find... 

I think it would be possible to replace the "link all" button with an "unlink all" button if all linked files are linked. 

image

GaetanGrd:


Zagrios:
I pushed a small fix to keep the shared-folders bulk button disabled while link or unlink actions are still running. This prevents a quick second click from queuing the opposite action and undoing the requested bulk operation.

Zagrios (approved):
Looks good after checking the relevant shared-folder changes and CI.

PR #1066, opened 2026-07-09
Add OpenXR runtime mismatch

### LINKED ISSUE #978: [FEAT.] : Report current active VR runtime (Steam or Meta)
### Problem

One of the most common issues encountered that is being reported in the server is XR runtime error from incorrectly configured setups, having an inbuilt checker to notify users would serve to alleviate many of the repetitive questions

### Solution

Have BSManager check the current active runtime before the user launcher beat saber and alert if the selected runtime does not match the install method of the current beat saber version, (or potentially setting it at the request of the user depending, if you think its doable I can add more steps to the workflow to account for this)

Considerations:

Windows (only popup, not setting automatically):

a. click "launch"  >runtime check is performed > send popup if mismatch identified > user must interact with the popup, either to proceed despite the error in the event that it is intended for some reason or to go and make the fix before coming back

b. selecting a BS install from the sidebar > runtime check is performed > send a popup if mismatch identified. Both should have a "do not remind me again" that sets a toggle in the settings

I dont think Linux needs any consideration for this, as it depends on the launch parameters that are explicitly set

### Alternative Solution

_No response_

### Operating System

None

### Additional Context

_No response_

GaetanGrd:


Zagrios:
Pushed a small follow-up fix for OpenXR launch edge cases and added focused coverage.

Zagrios:
Thanks @GaetanGrd for your contribution!

PR #1068, opened 2026-07-10
[fix-1067] remove skip steam launch option button

### LINKED ISSUE #1067: Request: remove "skip steam" button from #664
### Issue encountered

As discussion in the discord, further elaboration below:

1. Automatic steamvr launch no longer happens which was why I added it in the first place. I also believe that the behaviour change was correlation not causation due to the openxr updates around the time so the fix might not have been valid in the first place. 
2. This seems to cause confusion from users
3. Steam is required to be open for Steam DRM if the game is purchased from there so it needs to be open at some point during the launch, so useless outside of some niche debugging purposes that I havent been able to think of

Based on this I am suggesting the removal of the button, I can submit the patch to restore / clean up the code

### Expected Behavior

_No response_

### Steps to Reproduce

_No response_

### Operating System

None

### Version

_No response_

### Additional Context

_No response_

silentrald:
Close #1067


Zagrios:
Pushed a small follow-up fix to preserve existing Skip Steam shortcuts and saved launch behavior.

Zagrios (approved):
[BSM-AUTOPILOT-v4 repo=Zagrios/bs-manager pr=1068 generation=1 head=a6fc88bca6b5706ea86e75db7884e7aed96dae0c kind=approval] Exact candidate verified; checks green; no open material findings.

Zagrios:
Thanks @silentrald for your contribution!

PR #1074, opened 2026-07-16
Removing excess spaces, makes tab vs spaces consistent

LiamillionSS:
Reverting the automatic formatting changes before being able to save annoyed me, so instead of turning it off I went through every file and let it make the changes. Only things that were affected were tab vs spaces in some files, random spaces after brackets, missing or excess newlines at the end of the files, etc. 

Result is consistent and semantically correct formatting across the whole project. No changes were made to the actual code content, only cosmetic. No tests should be required but I built it locally just in case


Zagrios (approved):

Zagrios:
Thanks @LiamillionSS ❤️ Merged 🚀

PR #1104, opened 2026-08-25
[chore]: Update Russian translation

Dead0Duck:
## Scope

Corrections to the Russian translation. Translations of untranslated lines, correction of incorrect or machine-readable lines.

## Implementation

1. The first 4 lines regarding Refresh were updated. They used Unicode-encoded words, which made reading the file inconvenient; I also refined the translation itself.
2. Updated `generic.env.parse` because it used different wording, which could confuse users.
3. Fixed `see-changelog` used an inaccurate machine translation.
4. Refined `launch-mods.oculus-description` to improve clarity.
5. Fixed `launch-mods.proton-logs` and `launch-mods.proton-logs-description`—these strings were machine-translated, which had also incorrectly translated the word "Proton" itself.
6. Refined `launch-mods.parallel-views` to fix another machine translation issue.
7. Standardized `notes-per-second`: previously, NPS was fully written out as "Notes per Second" in some places. Now, we use the term consistently everywhere and include a clarification in parentheses.
8. Updated the names of other languages.
9. Numerous other minor fixes and tweaks.

## Emoji Guide

> 🟢 Nice refactor!


Zagrios:
Thanks @Dead0Duck ❤️ Merged 🚀

